PR #8, opened 2016-08-05
Only delete expired temporary bans, not perma bans

md678685:
Make sure that you're only deleting temporary bans and not permanent ones - permanent bans will always be considered "expired" since they always use an "end" value of -1, so the current query will always delete them.


DevLeoko:
Ya you are right thanks alot this would be a critical fail.
But your merege-request would only solve it for TempBans, other temporary punishments like tempMutes or tempWarns would still be deletet.

Thanks anyway I'll  fix that asap

md678685:
Sorry, didn't think of other temporary punishments. To be honest, we don't use warns or permanent mutes on our server, so it didn't occur to me to include them.

PR #20, opened 2016-11-29
added issue template

dutchy1001:
Added new 'issue template' to get quicker info to solve an issue.


dutchy1001:
the form might need tweaking to make it better, but atleast you will get some info to start with what makes life a lot easier.

PR #30, opened 2017-01-05
Fix for MySQL Port Changing Main connect method

ZombieHDGaming:
Noticed an error in the main MySQL connect method, where it doesn't add support for a custom port until the reconnect method, and since we don't want people complaining, figured I should add the quick fix and add the custom port in.


DevLeoko:
Thanks for your afford but you would also have to add the "Port: 3306"- Setting in the MySQL.yml

Also I've already added this in my last commit ;)
https://github.com/DevLeoko/AdvancedBan/commit/5454eaa6ab99b1e41b84a19caa39c84aa266e35b

Issue #44, opened 2017-01-27
Tab completion for nickname


Twonox:
Ab version: latest
Serveur version: latest 1.8.8 (with bungeecord)

When I want to complete a nick, the wrote chars are reset and all online player's name is browsed by tabulation, no completion for target nick.

e.g. I want to ban Twonox:
1. I write /ban Two
2. I use [TAB] to complete the nick
3. TWO is replaced by AAA, then by BBB etc.

ZombieHDGaming:
Since someone will eventually ask you this, are you using this on Bukkit/Spigot or BungeeCord, because your server version is a bit confusing.

If you are using Bukkit/Spigot, I will look into this and try to find the issue and make a pull request. If you are on BungeeCord, I will still try and find the issue, but I may not be able to help you as I am not that experienced with BC. I would also check your server settings, and plugins, for anything to do with tab completion, as this may also cause an issue.

md678685:
@ZombieHDGaming I assumed he was running AB on the server which was behind BungeeCord. 
In Bukkit and Spigot it is impossible to globally override tab completion for all commands; plugins can only do it per-command and there are no "settings" for it. BungeeCord usually passes tab completion through to the server unless it is disabled, in which case it won't work at all.

Regardless, I can confirm that this is an issue on Paper 1.11.2, which also doesn't alter tab completion behaviour as far as I know.

@Twonox Could you paste a list of plugins you are running?

Twonox:
I use Ab with bungee, not in single server.

Plugins: https://hastebin.com/gomaxusawo.nginx

md678685:
What Bungee plugins are you running?

Twonox:
AdvancedBan, BungeeAuthmeBridge, BungeeMotd, Bungeeperms, CommandDispatcher, SkinRestorer and Socket4MC

ZombieHDGaming:
@md678685 Actually, there is an event for tab completion to prevent it, and edit the completions, I am currently testing other commands for custom completions. Also, your bug confirmation is actually null. I just tested AdvancedBan 2.0.6 on the latest Spigot and PaperSpigot, freshly built, and tab complete works fine for players. As for you @Twonox, this is an interesting situation. I have not tested BC AdvancedBan yet, but I will have to test it and see if I get the same result as you do.

So yes. Current Standing: Spigot/PaperSpigot/Bukkit are fine, BungeeCord is still currently unknown. I think this is safe to say, this is a per-server error basis.

[...]

PR #46, opened 2017-01-29
Tab Complete Rewrite

ZombieHDGaming:
As it has been noted in issue #44, BungeeCord versions are having issues tab completing player names. While I have very little development skills for BungeeCord, theoretically, this should fix the issue and allow players to be tab completed based on text already given, as long as there is only 1 argument. A dev or Leoko can comment on other fixes for this, but I figured it was a quick fix, so why not. :)


DevLeoko:
Did you develop this in your browser or windows-texteditor? :D 

Nevermind fixed the code & thanks anyways for helping ^^

PR #88, opened 2017-06-01
FIX Warning MYSQL SSL CERT

jeremy100:
**Error on BungeeCord & Spigot using MySQL DB:**
"For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification."

**I have changed this params for temporal fix:**
"verifyServerCertificate" property is set to 'false'.
"useSSL" property is set to 'false'.

You can add "useSSL" property to the config of plugin for user's use a SSL Cert.

Please update ASAP!
Thanks

### DIFF
--- a/src/me/leoko/advancedban/manager/MySQLManager.java
+++ b/src/me/leoko/advancedban/manager/MySQLManager.java
@@ -89,7 +89,7 @@ public class MySQLManager {
 	
 	private void connect(){
 		try{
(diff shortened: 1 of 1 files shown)


ItzSomebody:
I was wondering why I kept getting this error. lol

PR #132, opened 2017-08-12
RedisBungee

Beelzebu:
This _should_ re implement all the features related to redisbungee of previous commits in the new version of advancedban (is was working since the last 4 hours, so anyone shouldn't have problems), also fix a bug with notifications and now if the server uses redis it is shown in the /advancedban command.
For now the IP of the player isn't sent through redisbungee for the /check command, when I added this it was causing a delay in the PreLoginEvent, I gonna find a way to make this safest, may be sending when the command /check is called, I don't know.


src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
```diff
@@ -144,12 +147,23 @@ public void sendMessage(Object player, String msg) {
 
     @Override
     public boolean hasPerms(Object player, String perms) {
-        return ((CommandSender) player).hasPermission(perms);
+        try {
+            return ((CommandSender) player).hasPermission(perms);
+        } catch (Exception ex) {
```
DevLeoko: How can this get an exception?

src/main/java/me/leoko/advancedban/bungee/listener/PubSubMessageListener.java
```diff
@@ -0,0 +1,54 @@
...
+                if (ProxyServer.getInstance().getPlayer(msg[1]) != null) {
+                    mi.kickPlayer(ProxyServer.getInstance().getPlayer(msg[1]), e.getMessage().substring((msg[0] + msg[1]).length() + 2));
+                }
+            } else if (e.getMessage().startsWith("notification ")) {
+                for (ProxiedPlayer pp : ProxyServer.getInstance().getPlayers()) {
+                    if (mi.hasPerms(pp, perm)) {
+                        mi.sendMessage(pp, e.getMessage().substring(13));
+                    }
+                }
+            } else if (e.getMessage().equals("refresh")) {
```
DevLeoko: Is that message send anywhere?

src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
```diff
@@ -35,10 +35,22 @@ public void setup() {
...
     }
+    
+    public void refresh() {
+        MethodInterface mi = Universal.get().getMethods();
+        mi.runAsync(() -> {
+            punishments.clear();
+            history.clear();
+            for (Object player : mi.getOnlinePlayers()) {
+                String name = mi.getName(player).toLowerCase();
+                load(name, UUIDManager.get().getUUID(name), mi.getIP(player));
```
DevLeoko: the load-Method returns an IterimData-Object, you will need to call the .accept() method so that the IterimData merges with the punishments and history list

src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
Beelzebu: The redisbungee PubSubListener only checks if the player has the perms for all bungee servers, so if the player isn't in the bungee server that is requesting this it gonna return a null.

src/main/java/me/leoko/advancedban/bungee/listener/PubSubMessageListener.java
Beelzebu: Nope, fixed.

src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
Beelzebu: Done.

Beelzebu:
I forgot to mention that now jars that are shaded within the final jar are minimized, so now the final jar is more lightweight (270Kb from 1.9 Mb).

src/main/java/me/leoko/advancedban/bungee/listener/PubSubMessageListener.java
DevLeoko: As the punishments are refreshed anyway when the player joins the server I don't get the point of reloading them for every bungee for every player. I'd consider just removing that refresh thing

src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
DevLeoko: Okay then add a not null check here or in the PubSubListener but just catching **all** exceptions is not a good solution for this ;)

PR #133, opened 2017-08-15
Notifications bug and jar file

Beelzebu:
The notifications wasn't showing correctly (I'm sorry :disappointed:) and now the final jar is lighter.


DevLeoko:
The minimize jar thing breaks the API for HSQLDB that's why I removed it

PR #136, opened 2017-08-26
%ID% - Variable in Layouts.yml and messages.yml

ItzSomebody:
Just so you can use %ID% in the Layouts.yml and in the messages.yml


ItzSomebody:
Actually hold on, I am working on some more code

PR #142, opened 2017-09-18
Update to latest Variables.

ItzSomebody:


DevLeoko:
But next time please use development, as the master contains the currently released version ;)

PR #174, opened 2018-02-03
Modified ConnectionListenerBungee to run async

ThePineapple27:
This prevents the thread from being blocked while performing long tasks, such as SQL queries or UUID lookups.

There is a wiki on this: https://www.spigotmc.org/wiki/common-development-pitfalls-bungeecord/#blocking-the-i-o-threads


DevLeoko:
Did you test if this works fine?

ThePineapple27:
Yup! ;)

PR #190, opened 2018-03-18
Allow for MariaDB Driver

rouing:
https://mariadb.com/resources/blog/mariadb-java-connector-driver-performance

The performance improvement may not actually be that much for something
as simple as a Ban plugin, but it is nice to have it as an option for
far reaching servers that have the SQL server across the world.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,11 @@
             <version>5.0.0-RC2</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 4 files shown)


Beelzebu:
The changes should be made in the development branch

rouing:
Whoops. I thought I was doing this in dev branch. Fixing.

PR #191, opened 2018-03-19
Added an option for making punishments silent with plugin messaging.

MelonCola:
Requested by a user of my ban GUI plugin.


SpooderCraft:
Yes we need that

DevLeoko:
This would cause NPEs if there is no silent argument

src/main/java/me/leoko/advancedban/bungee/listener/InternalListener.java
```diff
@@ -62,7 +62,7 @@ public void onPluginMessageEvent(PluginMessageEvent e) {
                             TimeManager.getTime() + punishment.get("end").getAsLong(),
                             punishment.get("calculation") != null ? punishment.get("calculation").getAsString() : null,
                             -1
-                    ).create();
+                    ).create(punishment.get("silent").getAsBoolean());
```
DevLeoko: Add a not-null check right here

MelonCola:
Edited.

Issue #79, opened 2017-05-18
[Suggestion] Disable some AdvancedBan commands


OrlandoLaCrue:
I would like some commands for AdvancedBan to be disable via config.yml

-Thanks!

A248:
Commands aren't dynamically registered on Spigot. Once defined in AdvancedBan's plugin.yml, a command is necessarily registered.

If you want to disable a command, you have a few options:

1. Remove the command's permission from your administrators. If they run the command, they'll receive a no-permission message (configured in the messages.yml). This is definitely the simplest solution.
2. Override the command in commands.yml. This will completely prevent the command from being registered by any plugin. Examples and information on how to do this are [here](https://bukkit.gamepedia.com/Commands.yml#Alias_Examples)

PR #201, opened 2018-04-28
MAYBE A solution of one of the feature requests

Test83318:
**MAYBE** This is a solution of one of the features request(#79).
To add a section in config.yml with *SOME* commands and to in Universal to check
if isn't config at "" to register the command as has been wrote in config.
If don't works I hope that the idea works to solving that "PROBLEM"

### DIFF
--- a/src/main/java/me/leoko/advancedban/Universal.java
+++ b/src/main/java/me/leoko/advancedban/Universal.java
@@ -57,16 +57,15 @@ public class Universal {
 
         mi.setCommandExecutor("advancedban");
         mi.setCommandExecutor("change-reason");
-        mi.setCommandExecutor("ban");
-        mi.setCommandExecutor("tempban");
-        mi.setCommandExecutor("ipban");
-        mi.setCommandExecutor("ban-ip");
-        mi.setCommandExecutor("banip");
-        mi.setCommandExecutor("tempipban");
-        mi.setCommandExecutor("tipban");
-        mi.setCommandExecutor("banip");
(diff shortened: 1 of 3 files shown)


src/main/java/me/leoko/advancedban/Universal.java
```diff
@@ -57,16 +57,15 @@ public void setup(MethodInterface mi) {
...
-        mi.setCommandExecutor("tempban");
-        mi.setCommandExecutor("ipban");
-        mi.setCommandExecutor("ban-ip");
-        mi.setCommandExecutor("banip");
-        mi.setCommandExecutor("tempipban");
-        mi.setCommandExecutor("tipban");
-        mi.setCommandExecutor("banip");
-        mi.setCommandExecutor("kick");
-        mi.setCommandExecutor("warn");
+        if(!config.getstring(Commands.ban).equals""){mi.setCommandExecutor(config.getstring(commands.ban));}
```
supertassu: That's not even valid Java syntax. You've created the modifications in GitHub's web UI. Have you tested them?

src/main/resources/plugin.yml
```diff
@@ -7,23 +7,23 @@ commands:
   AdvancedBan:
     description: Information about the Plugin
   ban:
-    description: Ban a user permanently
+    description: Ban a user permanently / CAN BE CHANGED
```
supertassu: They can't on Bukkit.

Test83318:
I've saw some cases of that, I don't know how was coded.
No, I didn't test them

PR #209, opened 2018-05-25
Update DatabaseManager.java

Baw-Appie:
Fixed that Korean is not displayed properly


DevLeoko:
Thanks for contributing :) have you tested these changes?

Baw-Appie:
Yes :)

DevLeoko:
Just noticed that this branch is not from the dev-branch.
Contributions have to be on the development branch sorry

DevLeoko:
@md678685 oh I thought it would screw up the development branch as I can not change the source branch only the target. But it makes sense that it does not as there are no changes in the master that are not in the dev just the other way around it would not work :D

Issue #214, opened 2018-06-12
Use consistent IDs in Punishments and PunishmentHisotry


BrainStone:
I think it would be better if the table `Punishments` and `PunishmentHisotry` shared the same IDs. When using punishments other than kick, they stay in sync anyways. So the auto index of `Punishments` should just be bumped when a kick is performed to make sure the tables stay in sync. This is especially important when modifying punishments or when using the `PunishmentHistory` to display the IDs of the bans so users can use it to ask to be pardoned.

BrainStone:
Closed by #216

PR #219, opened 2018-06-13
Sync auto IDs on every start

BrainStone:
This goes towards #214 too.

Also partially needed for migration from HSQL to MySQL.


BrainStone:
Don't merge it yet. It's not working.

BrainStone:
Now it's ready to be merged

PR #240, opened 2018-08-01
Update to 1.13 and fix api endpoint for determining the country for an IP address 

frost-byte:
Updated a few dependencies, bumped the version and fixed Country lookup by IP

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>me.leoko.advancedban</groupId>
     <artifactId>AdvancedBan</artifactId>
-    <version>2.1.7-${build}</version>
+    <version>2.1.6-${build}</version>
     <packaging>jar</packaging>
     <properties>
         <build>SNAPSHOT</build>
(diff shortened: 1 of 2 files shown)


src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
```diff
@@ -75,12 +76,15 @@ public String getFromUrlJson(String url, String key) {
             JSONParser jp = new JSONParser();
             JSONObject json = (JSONObject) jp.parse(new InputStreamReader(request.getInputStream()));
 
-            String[] keys = key.split("\\|");
```
DevLeoko: This code enabled the user to define a key which points to a child JsonObject.
Like result|playerdata|uuid
would point to result: { playerdata: { uuid: VALUE}}
Is there any reason for you to remove it? ^^

pom.xml
```diff
@@ -5,7 +5,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>me.leoko.advancedban</groupId>
     <artifactId>AdvancedBan</artifactId>
-    <version>2.1.6-${build}</version>
```
DevLeoko: Keep the version an 2.1.6 as it's not released yet

frost-byte:
I fixed those issues.  Sorry about the mix up with version and the mix up with getUrlFromJson.

DevLeoko:
You added two repos and one dependency but you don't seem to actually ever use them.

frost-byte:
I was unable to find the newer version of RedisBungee at the repository you were using.
The repo at [ci.lucko.me](https://ci.lucko.me/plugin/repository/everything/com/imaginarycode/minecraft/RedisBungee/) has a build of v0.4

If you'd like, I'll just set it back to 0.3.6-SNAPSHOT and remove the above repo.

I'll remove the repo/dependency for net.sourceforge.nekohtml.

frost-byte:
You're welcome. Thank you for merging it.

PR #266, opened 2018-12-09
Check for mute before passing to other chat plugins

jascotty2:
EventPriority.LOW / EventPriority.LOWEST processes before EventPriority.HIGH / EventPriority.NORMAL


DevLeoko:
This is indeed intended the way it is. As AdvancedBan should have the last word in who is able to use the chat an who not. Other plugins which are interested in whether a user is muted, should use the MONITOR EventPriority.

jascotty2:
While I don't disagree that you would want a mute to have the final say and whatnot, I believe you are being a tad paranoid about other plugins having poor behavior and un-muting players.

Checking for mute earlier in the event chain can eliminate needless processing related to chat formatting.

PR #302, opened 2019-12-28
Synchronise Connection Access

### LINKED ISSUE #203: Statement in the database error

 §8[]=====[§7Enabling AdvancedBan§8]=====[]
§8| §cInformation:
§8|   §cName: §7AdvancedBan
§8|   §cDeveloper: §7Leoko
§8|   §cVersion: §72.1.5-SNAPSHOT
§8|   §cStorage: §7MySQL (external)
§8| §cSupport:
§8|   §cSkype: §7Leoko33
§8|   §cMail: §7Leoko4433@gmail.com
§8|   §cGithub: §7https://github.com/DevLeoko/AdvancedBan/issues
§8|   §cDiscord: §7https://discord.gg/ycDG6rS
§8| §cUpdate:
§8|   §7There is a new version available! [2.1.4]
§8[]================================[]§f
 
 Bungeecord
 
??9This server is running BungeeCord version git:BungeeCord-Bootstrap:1.12-SNAPSHOT:fde2c3f:1310 by md_5

_Please povide the EXACT steps required to reproduce the problem..._
1.
2.
3.

_Any error/log post it through [pastebin.com](http://pastebin.com) and link it here._

https://pastebin.com/QQNkj8MF

Latest.log:
https://pastebin.com/k8qvn0XR

A248:
**What we know about** #203 

- Happens only with MySQL external storage mode
- Does not happen on every player login.

**Why Synchronise**

`Connection` objects are not thread-safe per multiple StackOverflow answers. Synchronisation is very simple yet very important. Even if this is not the fix to #203 , it is still good practice to use mutual exclusion in multi-threaded environments.

According to several StackOverflow answers, `Connection` objects are not thread-safe for MySQL JDBC (but possibly thread-safe in later versions). According to another user, connections are thread-safe if the driver is compliant with the SQL standard. MySQL does not comply with the standard, which is likely why their connection is not thread-safe. HSQLDB touts full compliance. This would explain why #203 only occurs with MySQL.

**Why This May Not Fix 203**
Concurrency problems may arise if multiple threads access a single state simultaneously. AsyncPlayerPreLoginEvent is asynchronous and may thus give rise to such errors. However, as a prerequisite, multiple threads must access the same state very quickly. I have encountered #203 in environments where it did not seem that the connection object would be used often.

Nevertheless, synchronisation should be performed when accessing JDBC `Connection` objects, even if the synchronisation does not fix #203 .

### DIFF
--- a/src/main/java/me/leoko/advancedban/manager/DatabaseManager.java
+++ b/src/main/java/me/leoko/advancedban/manager/DatabaseManager.java
@@ -18,6 +18,8 @@ import java.sql.*;
  */
 public class DatabaseManager {
 
+	private final Object lock = new Object();
+	
     private String ip;
     private String dbName;
     private String usrName;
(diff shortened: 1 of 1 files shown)


A248:
That was my initial instinct; however, I presumed that would block access to the DatabaseManager's other methods.

A248:
Nevermind; did more research on the topic. Synchronized keyword is favourable.

DevLeoko:
Thanks for the contribution ;D

PR #309, opened 2020-03-13
Add support for old notification permissions

A248:
Ensures the notification permission format, ab.TYPE.notify and ab.TYPE.undoNotify, continue to work on older server configurations. _TYPE_ is any `PunishmentType`.

To avoid breaking changes to MethodInterface calls, MethodInterface#notify(String, List) is retained. MI implementations, of course, must unavoidably be updated to support #notify(String, String, List).

Every effort is made to ensure that, if a player has both permissions from both formats, he or she will not receive two notification messages. For example, if a player has permission `ab.ban.notify` and `ab.notify.ban`, the player should not receive two messages. However, this behaviour, is not guaranteed in all MI implementations (on BungeeCord + Redis, the message would be sent twice).

### DIFF
--- a/src/main/java/me/leoko/advancedban/MethodInterface.java
+++ b/src/main/java/me/leoko/advancedban/MethodInterface.java
@@ -400,8 +400,22 @@ public interface MethodInterface {
      * @param perm         the perm
      * @param notification the notification
      */
-    void notify(String perm, List<String> notification);
-
+    default void notify(String perm, List<String> notification) {
+    	notify(perm, null, notification);
+    }
+    
+    /**
+     * Broadcast a message to every user with the given perm OR the alt perm.
+     * Effort is made to ensure that the same message is not sent twice
+     * if the user has both permissions, but this is not guaranteed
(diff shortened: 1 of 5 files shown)


DevLeoko:
This pull request solves the issue that we dropped support for the old permission nodes but introduces the (imho more severe) bug where all Staff on a Redis-Network with *-Perms will receive duplicate messages.

Also, `void notify(String perm, String altPerm, List<String> notification);` seems like a bad/unmodular method signature to me. As the fact that there is exactly one alternative permission is completely arbitrary. (Would need a new method if there ever were to be two alt perms).

Probably the right way to do it would be to pass an array with permission nodes. But as the notify permission is not critical at all, I'd vote to just leave it as it is.

A248:
If you suppose. I would prefer to maintain support for older permission nodes, but it's your call.

DevLeoko:
I really appreciate your contribution but I really don't want to accept PRs that knowingly introduce new bugs

PR #315, opened 2020-04-05
A bans plugins should respond to events first, so we use lowest priority

BillyGalbreath:


DevLeoko:
That's not necessarily true.
The decision about whether a player is banned should be final and not changed.
We already changed the priority to HIGH (in the dev branch) so other plugins can intervene if they really need to.
Also, the dev version checks whether the player has already been denied by another plugin so bot attacks can be blocked and don't crash the server because AdvancedBan checks their ban status.

On a side note: please use the dev branch for such changes and resolve any merge conflicts

PR #316, opened 2020-04-05
Register permissions with commands

BillyGalbreath:
So the /help command isn't cluttered with commands players don't have permissions to use them


DevLeoko:
Oh ya, that's a good point.
Have you checked whether the AdvancedBan command handler is still called even if the player does not have the permission so that AdvancedBan can display its own no permission message?

Also please merge this change into the dev branch and not the master (and maybe merge the dev branch first into your branch as it seems outdated)

DevLeoko:
Closed due to no response

Issue #322, opened 2020-04-18
Intern UUID is always null


james090500:
Intern UUID fetcher will always return null. As the login event is called from a "LoginEvent" the player hasn't yet been added to the BungeeCord player list.

When attempting to join and AdvancedBans calls getInternUUID here:
https://github.com/DevLeoko/AdvancedBan/blob/master/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java#L238

It first tries to get the ProxiedPlayer. This will always return null as the player is still a PendingConnection at that point. This is causing #317 and a bug with a plugin i'm creating as Fake usernames with Spoofed UUID's aren't being picked up by external REST API's.

Issue #320, opened 2020-04-17
Hikari library, mysql connection improvement


tobgen:
_What version of AdvancedBan (`/AdvancedBan`) are you using?_
 2.1.8
 
_What kind of server do you have (`Bungeecord/single server`)?_
 Bungee + 5 spigot
 
_What server version (/version) are you using?_
1.15.2

_Please provide the EXACT steps required to reproduce the problem..._
1. Install AdvancedBan
2. Configure it for MySQL
3. Have the servers idle
4. Player logs in

_Any error/log post it through [pastebin.com](http://pastebin.com) and link it here._
https://paste.md-5.net/vugobumalo.sql

_Add any additional information below._
This is kinda expected if your server has been idle for the timeout value.
Hikari can solve this by keeping the connection to the database alive.
Just a thought.

DevLeoko:
We have set "autoReconnect" to true, so in theory, it should reconnect once the connection is gone. But yes it would be a good idea to implement/use some more sophisticated MySQL (Pool) connection management. Thanks for pointing that issue out

tobgen:
Reconnect works and does what it is suppose to do , so the issue is just "cosmetic" in the logs.

DevLeoko:
The latest version of AdvancedBan now uses Hikari connection pooling. Thanks for the feature request 👍 
https://www.spigotmc.org/resources/advancedban.8695/update?update=335170

PR #323, opened 2020-04-19
Fix for Intern UUID fetcher returning null on join when UUID not cached

james090500:
Fixes issue #322

### DIFF
--- a/src/main/java/me/leoko/advancedban/Universal.java
+++ b/src/main/java/me/leoko/advancedban/Universal.java
@@ -248,12 +248,11 @@ public class Universal {
      *
      * @param name the name
      * @param ip   the ip
-     * @param fullUUID the uuid
      * @return the string
      */
-    public String callConnection(String name, String ip, UUID fullUUID) {
+    public String callConnection(String name, String ip) {
         name = name.toLowerCase();
-        String uuid = fullUUID.toString().replace("-", "");
+        String uuid = UUIDManager.get().getUUID(name);
         if (uuid == null) {
             return "[AdvancedBan] Failed to fetch your UUID";
         }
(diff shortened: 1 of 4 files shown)


DevLeoko:
_fullUUID.toString().replace("-", "");_ introduces UUID processing logic into the Universal class.
All UUID processing logic should be in the UUIDManager, as any other class should not make any assumption about UUIDs other than that they are unique strings used to identify a player.

So the Universal.java class assuming that "-" have to be removed in an issue.

Also, this causes a bug when the UUID fetcher is disabled. (Again this would not happen if UUID logic would be kept in the UUID Manager).

Two possible solutions:
1)
Make a function in the UUIDManager to add UUIDs to the activeUUIDs cache which can be called from the connection listeners or Universal.

2\)
Add a function getUUID(String name, String internalUUID) and move the UUID processing logic to the UUIDManager.

james090500:
Add per requested. I have moved UUID logic to UUIDManaged and from the connection listeners add the uuid and name to cache.

james090500:
Fixed bungee import within UUID Manager

DevLeoko:
This still causes issues with disabled UUID fetcher.

Maybe rather make a method _supplyInternUUID(name, uuid)_ in the UUID Manager, which only adds that UUID to the cache if the mode is set to INTERN or MIXED

james090500:
As requested

A248:
My goodness @DevLeoko ! This looks like the fix to #203 , sitting before our very eyes!

DevLeoko:
No, I don't think that this is related to that issue.
I think all those issues are related to the poor database connection handling we do ^^
See: #320

A248:
It seemed to me rather that we weren't setting the UUID in the parameterised statement, but I could be wrong.

Issue #351, opened 2020-05-16
Dependency Relocation


A248:
**What version of AdvancedBan (`/AdvancedBan`) are you using?**
 2.1.9-RELEASE
 
**What kind of server do you have (`Bungeecord/single server`)?**
 Single
 
**What server version (/version) are you using?**
For the example,
`Paper version git-Paper-284 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)`

**Please provide the EXACT steps required to reproduce the problem.**
1. Install AdvancedBan
2. Install another plugin which uses dependencies from AdvancedBan
3. Use a server implementation which includes dependencies from AdvancedBan
4. View/monitor for conflicts

**Any error/log post it through [pastebin.com](http://pastebin.com) and link it here.**
Here is an example, using AntiVPN (which relies on hsqldb) with AdvancedBan, https://pastebin.com/1tAfmaK4, the main takeaway of which is
```
[AntiVPN] Loaded class org.hsqldb.jdbc.JDBCDriver from AdvancedBan v2.1.9 which is not a depend, softdepend or loadbefore of this plugin.
```

**Add any additional information below.**
If 2 plugins uses different versions of shaded but not relocated dependencies, linkage errors may occur. The above example is the best-case scenario. I would presume plugins may encounter NoSuchMethodError / NoClassDefFoundError during operation if there is a versioning mismatch. The CraftBukkit wiki mentions "unintended behavior or exceptions" (https://bukkit.gamepedia.com/Using_External_Libraries_with_Plugins#Shade_the_external_libraries_into_the_plugin.E2.80.99s_jar). slf4j, in particular, *should* fail fast rather than permit the problem to silently persist (see http://www.slf4j.org/codes.html#multiple_bindings).

I originally setup 2 examples – the one described above with AntiVPN, and another intended to produce a slf4j binding conflict. I used multiple versions of Paper with slf4j included, as well as a plugin with slf4j shaded. Oddly enough, even though I verified that multiple slf4j bindings were on the classpath, I was unable to create such a conflict. However, I did find an error report related to another plugin using slf4j, where the plugin included slf4j in its jar: https://forums.minehut.com/topic/7113-an-internal-error-occurred-while-trying-to-perform-this-command/ (Use CTRL+F and search for "slf4j").

**The Solution**
[...]

PR #352, opened 2020-05-17
Relocate dependencies using maven-shade

A248:
This PR is to solve #351 by relocating AdvancedBan's shaded dependencies to *me.leoko.advancedban.shaded.**

There's still 1 issue, which, although minor, I would prefer to solve.

~~**This is still in progress** for 2 reasons, which, although they are minor, I would prefer to solve.~~

### ~~Log Messages~~

~~For one, I haven't found a way to shorten these log messages, nor to get them to log at the correct level.~~

~~It is possible to use a log4j2.xml file and direct the server instance to use it. However, doing so only changes the output of the message before the ':', after which the message is determined by slf4j-simple and is thus constant.~~

### HSQLDB as a registered Driver

For two, other plugins still use AdvancedBan's hsqldb, even though it is relocated. I luckily realised this because I accidentally left AntiVPN on the test server, which uses sqlite.
```
[13:57:25 INFO]: [com.zaxxer.hikari.HikariDataSource] AntiVPN-SQLite - Starting...
[13:57:25 WARN]: [AntiVPN] Loaded class me.leoko.advancedban.shaded.org.hsqldb.jdbc.JDBCDriver from AdvancedBan v2.1.9 which is not a depend, softdepend or loadbefore of this plugin.
[13:57:25 INFO]: [com.zaxxer.hikari.HikariDataSource] AntiVPN-SQLite - Start completed.
```
At the very least, AntiVPN *is* using its own HikariCP. This is an improvement over #351 . 
However, HikariCP detects and uses AdvancedBan's relocated hsqldb. I am not sure if this a feature or a bug; it may be that this is the intended behavior of AntiVPN/HikariCP with sqlite/hsqldb.

I will link an [AntiVPN](https://github.com/egg82/AntiVPN) issue to this one in case its developer may know any way to solve this.

### DIFF
--- a/src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
+++ b/src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
@@ -22,8 +22,8 @@ public class DynamicDataSource {
             config.setUsername(usrName);
             config.setPassword(password);
         } else {
-        	// No need to worry about relocation because the maven-shade-plugin also changes strings
-        	String driverClassName = "org.hsqldb.jdbc.JDBCDriver";
+            // No need to worry about relocation because the maven-shade-plugin also changes strings
+            String driverClassName = "org.hsqldb.jdbc.JDBCDriver";
             Class.forName(driverClassName);
             config.setDriverClassName(driverClassName);
             config.setJdbcUrl("jdbc:hsqldb:file:" + mi.getDataFolder().getPath() + "/data/storage;hsqldb.lock_file=false");
--- a/src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
+++ b/src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
@@ -22,7 +22,10 @@ public class DynamicDataSource {
             config.setUsername(usrName);
             config.setPassword(password);
         } else {
-            Class.forName("org.hsqldb.jdbc.JDBCDriver");
+        	// No need to worry about relocation because the maven-shade-plugin also changes strings
+        	String driverClassName = "org.hsqldb.jdbc.JDBCDriver";
+            Class.forName(driverClassName);
+            config.setDriverClassName(driverClassName);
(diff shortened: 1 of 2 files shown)


A248:
Parallel issue https://github.com/egg82/AntiVPN/issues/65 on the AntiVPN github.

DevLeoko:
I think the way the messages are printed to the console is defined by the SimpleLogger.
If you really want to change that you should be able to just bind another logger:
http://www.slf4j.org/manual.html#swapping

A248:
Changing the binding to jdk14 cleared up the log messages and made them use the correct level :thumbsup:

```
[18:24:51 INFO]: [me.leoko.advancedban.shaded.com.zaxxer.hikari.HikariDataSource] HikariPool-1 - Starting..
```

A248:
I'm not sure why the checks were cancelled. It builds fine and passes unit tests on my local machine.

src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
```diff
@@ -22,7 +22,10 @@ public DynamicDataSource(boolean preferMySQL) throws ClassNotFoundException {
             config.setUsername(usrName);
             config.setPassword(password);
         } else {
-            Class.forName("org.hsqldb.jdbc.JDBCDriver");
+        	// No need to worry about relocation because the maven-shade-plugin also changes strings
+        	String driverClassName = "org.hsqldb.jdbc.JDBCDriver";
```
DevLeoko, line 26: Used TABs instead of spaces for code indentation

A248: Interesting. My IDE uses tabs by default. However, I was trying to copy the style of the surrounding code, which on my machine seems to use spaces.

I'll change to tabs nonetheless.

A248:
Yes, this can be merged. We've done all that can be expected of AdvancedBan with regards to relocating dependencies.

src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
A248: Note: Leoko and I realised spaces should have been used after all.

PR #360, opened 2020-05-29
Fixes UUID Cache not working for multi-case names

### LINKED ISSUE #317: Geyser Floodgate Support
Hey,
I am looking to use AdvancedBan on my Server, but I also want to use Geyser's FloodGate in conjunction. I've gotten everything to work but when logging in it throws a `[AdvancedBan] Failed to fetch your UUID`.
Would there be a workaround to this to make this operational to support both clients?

james090500:
if a player had multiple cases in their name the uuid cache would fail to work,

image

This fixes #317

### DIFF
--- a/src/main/java/me/leoko/advancedban/manager/UUIDManager.java
+++ b/src/main/java/me/leoko/advancedban/manager/UUIDManager.java
@@ -107,7 +107,7 @@ public class UUIDManager {
      */
     public void supplyInternUUID(String name, UUID uuid) {
         if (mode == FetcherMode.INTERN || mode == FetcherMode.MIXED) {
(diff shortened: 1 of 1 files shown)


DevLeoko:
For consistency, getInMemoryUUID should then also convert the name to lower case

A248:
`askAPI` should also use *toLowerCase*. Basically, any time a key is used on the map, it must be lower-cased.

james090500:
Requested changes

DevLeoko:
Thanks for contributing to AdvancedBan

benrobson:
So assumably if I was to pull the master of AdvancedBan and test this it should be operational with FloodGate?

neco222:
i still can't get it to work
How does this conversation work?
i can't understand

PR #361, opened 2020-06-01
Fix mute commands triggering false positives

### LINKED ISSUE #357: Player cannot use commands that start with any of the prohibbited commands
 2.1.9
 
Single
 
Paper-318

1. Mute player.
2. Player tries to use command starting with any of the prohibitted coommands. E.g there is an -"r" entry in a list of prohibitted commands.
3. Player cannot use commands starting with "r". E.g /region or /rtp

_Any error/log post it through [pastebin.com](http://pastebin.com) and link it here._

Problem is here: [https://github.com/DevLeoko/AdvancedBan/blob/652f414a545fe9b469f870cf53846deaeb871622/src/main/java/me/leoko/advancedban/Universal.java#L208](url)

A248:
Fixes #357

### DIFF
--- a/src/main/java/me/leoko/advancedban/Universal.java
+++ b/src/main/java/me/leoko/advancedban/Universal.java
@@ -200,16 +200,55 @@ public class Universal {
      * @return the boolean
      */
     public boolean isMuteCommand(String cmd) {
-        // Strip everything but the main command
-        cmd = cmd.split(" ")[0];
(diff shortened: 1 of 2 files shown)


DevLeoko:
This breaks support for sub-command mutes.
For example, you won't be able to mute a command like "/party msg"

A248:
Oh, I see. I'll fix that.

PR #362, opened 2020-06-08
Merge master to development

A248:
Will allow further progress on development without having to worry about conflicts with master. Yes, it happened again.

In the future, it may be best to prevent master getting ahead of the development branch, since this is somewhat annoying and slows things down. To this end, I suggest one of the following:

- As a simple solution, the default branch is changed to development. Releases will still be made based on the master branch, and the master branch will be considered more stable, just like now. However, because development would be the default branch, we would not have to worry about remembering to merge into the development branch.
- As another option, the idea of an unstable development branch and a stable master branch is scrapped entirely. The master branch (or the development branch, which ever name is preferred) would be used as the default branch, and it would be unstable. The stable version of AdvancedBan would still be found on the Spigot page as well as GH releases. Since Github tags the commit at which a release is made, it wouldn't be hard to view the 'stable' state of the codebase.

In any case, I think the default branch should definitely be the "unstable" branch, so we don't have to merge master into development when we forget to PR into development.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
     <version>2.1.8</version>
     <packaging>jar</packaging>
     <properties>
-        <build>SNAPSHOT</build>
+        <build>RELEASE</build>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <skipTests>false</skipTests>
         <trimStackTrace>true</trimStackTrace>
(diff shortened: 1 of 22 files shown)


DevLeoko:
I really want to keep the master branch as the default one as this is the branch shown to people visiting the GitHub Repo Page and I feel like the first thing they should see is the current stable version.
Also scraping the master and dev branch into one and using tags for stable builds comes with the issue that it would be hard to make hotfixes on the currently released version when there already have been other changes.

You still make some good points there but all in all, I think the best way to go here is just to double-check which branches we are using and decline unintended PRs into the master.

PR #373, opened 2020-06-26
Ensure AdvancedBan does not change foreign commands

A248:
AdvancedBan currently uses `Server#getPluginCommand`, which can return any PluginCommand, not just those belonging to the calling plugin. The effect of this is that AB accidentally sets the executor of other plugins' commands, supposing they define commands with the same name as those by AB.

### Example - Insights Plugin

Before:
* `/check` uses AB's /check
* `/insights:check` uses AdvancedBan's /check
* `/advancedban:check` reports blank output (This happens because the CommandExecutor is set to the JavaPlugin by default)
After:
* `/check` uses Insights's /check
* `/insights:check` uses Insight's /check
* `/advancedban:check` uses AB's /check

Ref: https://www.spigotmc.org/resources/insights-super-configurable-limits-a-sync-scans-realtime-1-8-1-16-1.56489/

### Solution

Use JavaPlugin#getCommand, which will use the correct command if multiple plugins define the same command.

### DIFF
--- a/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
+++ b/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
@@ -145,7 +145,7 @@ public class BukkitMethods implements MethodInterface {
 
     @Override
     public void setCommandExecutor(String cmd) {
-        PluginCommand command = Bukkit.getPluginCommand(cmd);
+        PluginCommand command = getPlugin().getCommand(cmd);
         if (command != null) {
             command.setExecutor(CommandReceiver.get());
         } else {
--- a/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
+++ b/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
@@ -145,7 +145,8 @@ public class BukkitMethods implements MethodInterface {
 
     @Override
     public void setCommandExecutor(String cmd) {
-        PluginCommand command = getPlugin().getCommand(cmd);
+        boolean friendly = getBoolean(getConfig(), "Friendly Register Commands", false);
+        PluginCommand command = (friendly) ? getPlugin().getCommand(cmd) : Bukkit.getPluginCommand(cmd);
         if (command != null) {
             command.setExecutor(CommandReceiver.get());
         } else {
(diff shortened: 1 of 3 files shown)


DevLeoko:
Is it possible that this would also prevent us from overriding Essentials punishment commands? If that's the case, we might have to come up with a completely other solution.

But yes currently AdvancedBan registers commands in the wrong way. Thank you for looking into this.

A248:
Yes, it would prevent that. Bukkit does not seem to want plugins to decide this behaviour.

There are a few options we can consider:

1. Tell users to use the commands.yml and add an override to use AdvancedBan's commands.
2. A config option which toggles the behaviour of this PR, between `getPluginCommand` and `getCommand`. It should be the latter by default, though, so that AB is friendly with other plugins.
3. Add `loadbefore` to AB's plugin.yml so that it loads before Essentials, and thus its commands take priority.
4. A config option which toggles a listener to use the `PlayerCommandPreprocessEvent` and intercept the command.

Let's eliminate No4 since it disables tab completion, is harder to work with, and really isn't the right way to listen for a command (even less so than `getPluginCommand`).

No3 might have some merit, but it introduces another dependency ordering. This would slightly reduce overall dependency flexibility on servers with a lot of plugins; it would not be possible to load a plugin after Essentials but before AdvancedBan. It also only works specifically for Essentials or any other plugin names which would need to be listed in the plugin.yml.

No#1 is in my opinion the best option, since it acts in accordance with how Bukkit was designed to handle conflicts. Once an entry is defined the commands.yml it is foolproof and not subject to any further plugin shenanigans.

No#2 might be what you're looking for, if you really need want overriding Essentials's commands to be as quick as possible. But again, this could lead to the same odd behaviour detailed in the above example.

DevLeoko:
Even though I see how it is bad practice, I think we should keep the default behavior to how it is.
Because it seems to work pretty well as we get nearly no complaints about unwanted command overrides. And a lot of servers use Essentials (or another plugin with some basic punishment system) and I am sure most of them will be confused why AdvacnedBan commands don't work anymore.

I guess the best way to go is with `getPluginCommand ` by default and have a list in the config for which commands to use `getCommand` instead to allow overrides

A248:
In that case, we can use a config option. However, I think it makes more sense and is more maintainable to have a single config on/off switch to use `getCommand`, instead of a list of commands.

The config option doesn't have to be very prominent since we've only had a 1 user so far with this problem, which is why a list is probably overkill.

DevLeoko:
Agreed

Issue #374, opened 2020-06-30
[Feature Request] Adding Notes to history without notifying player


zettlinger:
_What version of AdvancedBan (`/AdvancedBan`) are you using?_
 2.1.9
 
_What kind of server do you have (`Bungeecord/single server`)?_
 Waterfall
 
_What server version (/version) are you using?_
paper-24 (1.16)

_Add any additional information below._

i would like to be able to make notes into players History for other staff to see. For example how to deal with a player if they repeat offence or if player had agreed to some rules face to face with staff etc.
It would act like a warning without the drawbacks of adding a warning

Adding notes to history can streamline the punishment/rule keeping more consistent between diff staffs that catch someone

Issue #103, opened 2017-07-07
Request  - Leave a /note on a players history file.


SiuanSong:
Request
Ability to do /note player to leave a note on the players record without notifying them.  Could be useful for servers that have punishment commands that are not handled by AdvancedBan.  Such as GriefPreventions /softmute and /separate.   Also would be good for keeping track of players that toe the lines of the rules.

DevLeoko:
Ya that's a good idea will be added

SiuanSong:
Any update on this? :P

PR #376, opened 2020-07-06
note

tWizT3ddreaMr:
Addition of /note (name) (reason) /unote (id) and /notes (name)

#103 #374 (same)


src/main/java/me/leoko/advancedban/manager/UpdateManager.java
```diff
@@ -59,6 +59,7 @@ public void setup() {
             addMessage("UnBan:", "  Notification: \"&e&o%OPERATOR% &7unbanned &c&o%NAME%\"", 1);
             addMessage("UnMute:", "  Notification: \"&e&o%OPERATOR% &7unmuted &c&o%NAME%\"", 1);
             addMessage("UnWarn:", "  Notification: \"&e&o%OPERATOR% &7unwarned &c&o%NAME%\"", 1);
+            addMessage("UnNote:", "  Notification: \"&e&o%OPERATOR% &7unnoted &c&o%NAME%\"", 1);
```
DevLeoko, line 62: This is not how the UpdateManager works. Please take a look at it again and adjust your code. Feel free to ask if you need any help understanding how it works.

src/main/java/me/leoko/advancedban/utils/Command.java
```diff
@@ -264,6 +329,8 @@
                     MessageManager.sendMessage(sender, "Check.BanReason", false, "REASON", ban.getReason());
                 }
                 MessageManager.sendMessage(sender, "Check.Warn", false, "COUNT", PunishmentManager.get().getCurrentWarns(uuid) + "");
+
+                MessageManager.sendMessage(sender, "Check.Note", false, "COUNT", PunishmentManager.get().getCurrentNotes(uuid) + "");
```
DevLeoko, line 333: You forgot to add the corresponding message to the message file

src/main/java/me/leoko/advancedban/utils/Punishment.java
```diff
@@ -137,7 +137,7 @@ public void create(boolean silent) {
 
             if (getType().getBasic() == PunishmentType.BAN || getType() == PunishmentType.KICK) {
                 mi.runSync(() -> mi.kickPlayer(getName(), getLayoutBSN()));
-            } else {
+            } else if (getType().getBasic() != PunishmentType.NOTE){
```
DevLeoko, line 140: You just wanted to avoid the message which is good, but you accidentally also disabled the cache update this way

src/main/java/me/leoko/advancedban/utils/PunishmentType.java
```diff
@@ -11,6 +11,7 @@
     MUTE("Mute", null, false, "ab.mute.perma"),
     TEMP_MUTE("Tempmute", MUTE, true, "ab.mute.temp"),
     WARNING("Warn", null, false, "ab.warn.perma"),
+    NOTE("Note", null, false, "ab.Note"),
```
DevLeoko, line 14: I think the permission node would rather be something like "ab.note.use" according to our convention

src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
```diff
@@ -134,6 +134,9 @@ public void discard(String name) {
             try (ResultSet rs = DatabaseManager.get().executeResultStatement(current ? SQLQuery.SELECT_USER_PUNISHMENTS : SQLQuery.SELECT_USER_PUNISHMENTS_HISTORY, target)) {
                 while (rs.next()) {
                     Punishment punishment = getPunishmentFromResultSet(rs);
+                    if(put.getBasic()==PunishmentType.WARNING && ( put == punishment.getType().getBasic() || punishment.getType().getBasic() == PunishmentType.TEMP_WARNING) && (!current || !punishment.isExpired())) {
+                    	  ptList.add(punishment);
+                	}else
```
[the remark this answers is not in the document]
tWizT3ddreaMr: For the warns command it was sending a null for PunishmentType which would work with how bans work, cause you cant see your own active bans, but it doesn't work with the notes. Because it was returning the notes aswell. I figured to set them together, maybe I could set a another section to return the active warnings.

src/main/java/me/leoko/advancedban/utils/Punishment.java
tWizT3ddreaMr: Will revise that then. I didn't notice that.

src/main/java/me/leoko/advancedban/utils/PunishmentType.java
tWizT3ddreaMr: Will correct

src/main/java/me/leoko/advancedban/manager/UpdateManager.java
tWizT3ddreaMr: That was 100% my bad. Misunderstood what the class was. And didn't think twice because I reset the configs several times during my tests

src/main/java/me/leoko/advancedban/utils/Command.java
tWizT3ddreaMr: Will correct

src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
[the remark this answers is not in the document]
tWizT3ddreaMr: Didn't address any of this with my latest commit because I wanted to see what you say about it. My reasons were there. 
Also forgot a null check for put. So history throws a null pointer. But if this isn't the way to go would rather correct it correctly.

tWizT3ddreaMr:
Well the checks have failed for failing to transfer bstats. I haven't touched the pom.xml. Whenever you look at this pull again could you run it again, maybe their servers went down.

src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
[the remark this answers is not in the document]
DevLeoko: I am really confused as to why we used 'null' before in the warns command as that should have returned all punishments ^^

I'm still not 100% sure what you are trying to achieve with your code here. Just using 'PunishmentType.WARNING' should return you all the warnings and 'PunishmentType.NOTE' all the notes. Without the need for that additional if-statement.

Also notice how 'put.getBasic()' is the same as 'put' as you are required to use a basic type anyway for the put parameter (see java docs above)
And `punishment.getType().getBasic() == PunishmentType.TEMP_WARNING` always returns false as `TEMP_WARNING` is not the basic type. The basic type is `WARNING`.

DevLeoko:
Another two things I noticed:
1) Please double-check to use our default code intention (which I think is 4 spaces "    ") and not TABs
2) This should be merged into the dev branch as the master should always contains the currently released version

src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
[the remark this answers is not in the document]
tWizT3ddreaMr: That was the problem, I was assuming that it would return true on one or the other. That is why I did add that to keep the warns command working. But I should have checked what the getBasic did properly rather than making assumptions, which the getType should have clued me in that I should have checked.

Will revert.

DevLeoko:
Oh sorry, as the title of the last commit was "tabs" I thought, you only changed the TABs and not the other issue. Just realized you actually did ^^

So sorry for the delay and thanks a lot for contributing to AdvancedBan, I am sure our users are going to appreciate it 👍

A248:
Looking over this, I'm pretty sure the changes to `PunishmentType` is an API break - changing the ordering of enum constants changes their ordinal values.

To fix this, `NOTE` should be the last entry in the enum. This should be changed before development is merged into master. @DevLeoko

DevLeoko:
Thanks for noticing @A248

PR #380, opened 2020-07-25
Fix unnote usage message

JasonHorkles:
Fixes the unnote usage message saying unwarn instead of unnote


DevLeoko:
Thanks for the fix 👍

PR #388, opened 2020-08-03
Full Maven Update

JulianDev24:
Every package is splitted into maven modules.


A248:
This is an API breaking change.

DevLeoko:
> This is an API breaking change.

I'll make sure to communicate that when publishing the update

OopsieWoopsie:
I'd suggest to change the artifactId of bukkit and bungee modules, because its confusing to call them BukkitMain and BungeeMain, something like "AdvancedBan-Bukkit" and "AdvancedBan-Bungee" would be better.

A248:
There doesn't need to be a .gitignore in each module - you can simply have one in the root directory and use relative paths if necessary.

DevLeoko:
This PR introduces a lot of redundancies in the POMs and does not compile the original AB Version that works on all server implementations. Thus this PR will be closed.

Thanks though for initiating this process, we have implemented similar changes now available in the dev branch.

PR #393, opened 2020-08-08
Database queries need not block each other

A248:
This small change means AdvancedBan can actually execute database queries concurrently, truly fulfilling the use of HikariCP.


A248:
I do have a warning for this. Other parts of AdvancedBan are not thread safe although commands are run in an async task.

It may be that the DatabaseManager's synchronisation when executing queries is indirectly preventing a whole host of other thread safety problems. That is, if each thread executes database queries sequentially, due to the length of time it takes to execute a database query, this may have the effect preventing post-query logic from overlapping.

In the absence of synchronisation, this would not be the case.

A248:
Closing this for now

PR #399, opened 2020-08-16
VelocitySupport

fede1132:
I made AdvancedBan compatible with Velocity Powered.

List of things must be revised:
* ChatListenerVelocity.java
* ConnectionListenerVelocity.java

Known bugs:
* UUID Fetcher not working
* ~~Comments not saved into config file~~ Fixed
* ~~MySQL not working~~ (Working but plugin size is 7mb)

Things missing (due to Velocity "limitations")
* Plugin messaging

Thanks to:
@JavaFactoryDev for the storage library I used.

[Download beta version](https://drive.google.com/file/d/17D1J4IjJ-bXS9Pfyo1iw6pkv_PPgH-qb/view?usp=sharing)

### DIFF
--- a/core/src/main/resources/MySQL.yml
+++ b/core/src/main/resources/MySQL.yml
@@ -1,0 +1,7 @@
+MySQL:
+  IP: localhost
+  DB-Name: database-name
+  Username: admin
+  Password: superSecret
+  Port: 3306
+  Properties: 'verifyServerCertificate=false&useSSL=false&useUnicode=true&characterEncoding=utf8'
--- a/velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
+++ b/velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
@@ -11,41 +11,22 @@ import me.leoko.advancedban.Universal;
 import me.leoko.advancedban.velocity.listener.ChatListenerVelocity;
 import me.leoko.advancedban.velocity.listener.ConnectionListenerVelocity;
 
-import javax.net.ssl.HttpsURLConnection;
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.X509TrustManager;
 import java.io.*;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLClassLoader;
-import java.net.URLConnection;
-import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
-import java.sql.Driver;
-import java.sql.DriverManager;
-import java.sql.SQLException;
-import java.util.ArrayList;
 import java.util.logging.Logger;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipInputStream;
 
 @Plugin(
         id = VelocityMain.VelocityPluginInfo.PLUGIN_ID,
         name = VelocityMain.VelocityPluginInfo.PLUGIN_NAME,
-        version = VelocityMain.VelocityPluginInfo.PLUGIN_VERSION,
+        version = "${project.version}",
         authors = {"Leoko", "Fede1132(Added velocity support)"}
 )
 public class VelocityMain {
     public static class VelocityPluginInfo {
         public static final String PLUGIN_ID = "advancedban";
         public static final String PLUGIN_NAME = "AdvancedBan";
-        public static final String PLUGIN_VERSION = "2.2.1";
     }
     private static VelocityMain instance;
-    private ArrayList<File> loaded = new ArrayList<>();
     public final ProxyServer server;
     public final Logger logger;
     public final File folder;
--- a/velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
+++ b/velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
@@ -1,0 +1,76 @@
+package me.leoko.advancedban.velocity;
+
+import com.google.inject.Inject;
+import com.velocitypowered.api.event.Subscribe;
+import com.velocitypowered.api.event.proxy.ProxyInitializeEvent;
+import com.velocitypowered.api.event.proxy.ProxyShutdownEvent;
+import com.velocitypowered.api.plugin.Plugin;
+import com.velocitypowered.api.plugin.annotation.DataDirectory;
+import com.velocitypowered.api.proxy.ProxyServer;
+import me.leoko.advancedban.Universal;
+import me.leoko.advancedban.velocity.listener.ChatListenerVelocity;
+import me.leoko.advancedban.velocity.listener.ConnectionListenerVelocity;
+
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+import java.io.*;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.net.URLConnection;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.logging.Logger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+@Plugin(
+        id = VelocityMain.VelocityPluginInfo.PLUGIN_ID,
+        name = VelocityMain.VelocityPluginInfo.PLUGIN_NAME,
+        version = VelocityMain.VelocityPluginInfo.PLUGIN_VERSION,
+        authors = {"Leoko", "Fede1132(Added velocity support)"}
+)
+public class VelocityMain {
+    public static class VelocityPluginInfo {
+        public static final String PLUGIN_ID = "advancedban";
+        public static final String PLUGIN_NAME = "AdvancedBan";
+        public static final String PLUGIN_VERSION = "2.2.1";
+    }
+    private static VelocityMain instance;
+    private ArrayList<File> loaded = new ArrayList<>();
+    public final ProxyServer server;
+    public final Logger logger;
+    public final File folder;
+
+    @Inject
+    public VelocityMain(ProxyServer server, Logger logger, @DataDirectory final Path folder) {
+        instance = this;
+        this.server = server;
+        this.logger = logger;
+        this.folder = folder.toFile();
+    }
+
+    @Subscribe
+    public void onInitialize(ProxyInitializeEvent event) {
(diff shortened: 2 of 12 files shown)


DevLeoko:
Looks good all in all, but:
- The plugin version should not be a hardcoded constant in the code. If velocity does not support reading the version from the plugin.yml or bungee.yml then please try something like this: https://stackoverflow.com/questions/3697449/retrieve-version-from-maven-pom-xml-in-code/41791885
- Please try to do something about the plugin size, this would make AdvancedBan about 3x it's original size :(
- Why did you add the `MySQL.yml`? All MySQL Data are available in the config.yml
- Do you know why the UUID Fetcher is not working? Can you do something about that? Seems to me like a major issue

fede1132:
> Looks good all in all, but:
> 
> * The plugin version should not be a hardcoded constant in the code. If velocity does not support reading the version from the plugin.yml or bungee.yml then please try something like this: https://stackoverflow.com/questions/3697449/retrieve-version-from-maven-pom-xml-in-code/41791885
> * Please try to do something about the plugin size, this would make AdvancedBan about 3x it's original size :(
> * Why did you add the `MySQL.yml`? All MySQL Data are available in the config.yml
> * Do you know why the UUID Fetcher is not working? Can you do something about that? Seems to me like a major issue

1. Ok, I'm gonna do this
1. Forgot during some tests
1. I'm trying to understand this..

DevLeoko:
I don't really see why you use that particular storage library. If it's too big and Velocity don't have their own YAML parser then just use a different one like SnakeYAML which is just 300kb

JavaFactoryDev:
Yes and LightningStorage is only 291 kb. https://prnt.sc/u0m0wk
Just saw I got mentioned.

fede1132:
> I don't really see why you use that particular storage library. If it's too big and Velocity don't have their own YAML parser then just use a different one like SnakeYAML which is just 300kb

The file size is 7mb because Velocity does not have a built-in mysql-connector so I need to shade it into the jar.
And anyway Leonhard's library is the best library I've ever seen, it automatically reloads config when these are updated.

DevLeoko:
Oh sorry, my bad, the message above sounded a bit like the storage library was responsible for the large size.
In that case we should maybe not include the Velocity build in the bundled version but make it build separately.

Also, let me know when there are any updates on the UUID situation ^^

fede1132:
I think the problem Is with outed API urls. I'm trying to add Support for playerdb that Is a cached player API for Steam Games, Minecraft and Valve games. Powered by Steam

DevLeoko:
> In that case we should maybe not include the Velocity build in the bundled version but make it build separately.

Have you given that a thought?

fede1132:
> > In that case we should maybe not include the Velocity build in the bundled version but make it build separately.
> 
> Have you given that a thought?

I'm busy in these days I'll di this maybe this weekend

core/src/main/resources/MySQL.yml
```diff
@@ -0,0 +1,7 @@
+MySQL:
+  IP: localhost
```
A248, line 2: Why is the MySQL.yml brought back? I thought those settings were moved into config.yml

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,368 @@
...
+        } catch (Exception exc) {
+            return null;
+        }
+    }
+
+    @Override
+    public String getVersion() {
+        String version = "Unknown";
+        try (InputStream stream = getClass().getClassLoader().getResourceAsStream("velocity-plugin.json"); InputStreamReader read = new InputStreamReader(stream)) {
+            JsonObject obj = new JsonParser().parse(read).getAsJsonObject();
```
A248, line 80: There has to be a better way to do this. This method does IO and parses Json every time it is called. Why not use Velocity's methods on PluginContainer?

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,368 @@
...
+        messages = LightningBuilder.fromFile(messageFile).addInputStreamFromResource(messageFile.getName()).setConfigSettings(ConfigSettings.PRESERVE_COMMENTS).createYaml();
+        layouts = LightningBuilder.fromFile(layoutFile).addInputStreamFromResource(layoutFile.getName()).setConfigSettings(ConfigSettings.PRESERVE_COMMENTS).createYaml();
+        if (mysqlFile.exists()) {
+            mysql = LightningBuilder.fromFile(mysqlFile).setConfigSettings(ConfigSettings.PRESERVE_COMMENTS).createYaml();
+        } else mysql = config;
+    }
+
+    @Override
+    public String getFromUrlJson(String url, String key) {
+        try {
```
A248, line 57: Although I know this is somewhat outside the scope of this PR, this method's body is much duplicated. Rather than write it a 3rd time, perhaps there is a better way to consolidate all of it?

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
```diff
@@ -0,0 +1,57 @@
...
+
+@Plugin(
+        id = VelocityMain.VelocityPluginInfo.PLUGIN_ID,
+        name = VelocityMain.VelocityPluginInfo.PLUGIN_NAME,
+        version = "${project.version}",
+        authors = {"Leoko", "Fede1132(Added velocity support)"}
+)
+public class VelocityMain {
+    public static class VelocityPluginInfo {
+        public static final String PLUGIN_ID = "advancedban";
```
A248, line 26: Why are this class and its constants public?

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,368 @@
...
+    @Override
+    public String getInternUUID(String player) {
+        Optional<Player> optionalPlayer = getPlugin().server.getPlayer(player);
+        return optionalPlayer.map(value -> value.getUniqueId().toString().replace("-", "")).orElse(null);
+    }
+
+    @Override
+    public boolean hasPerms(Object player, String perms) {
+        if (player instanceof Player) {
+            return ((Player) player).hasPermission(perms);
```
A248, line 178: This is a completely unnecessary instanceof/cast. The way AdvancedBan is made, you need only `return ((CommandSource) player).hasPermission(perms)`

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    @Override
+    public boolean isOnline(String name) {
+        return getPlugin().server.getPlayer(name).isPresent();
+    }
+
+    @Override
+    public Object getPlayer(String name) {
+        return getPlugin().server.getPlayer(name).get();
```
A248, line 196: Using `Optional#get` changes the behaviour of this method from other MethodInterface implementations, making this one throw NoSuchElementException for nonexistent values. Also, the overridden signature can be changed to `Player`.

velocity/src/main/java/me/leoko/advancedban/velocity/listener/CommandReceiverVelocity.java
```diff
@@ -0,0 +1,38 @@
...
+            args[0] = (VelocityMain.get().server.getPlayer(args[0]).isPresent() ? VelocityMain.get().server.getPlayer(args[0]).get().getUsername() : args[0]);
+        }
+        CommandManager.get().onCommand(commandSource, cmd, args);
+    }
+
+    @Override
+    public List<String> suggest(CommandSource source, @NonNull String[] args) {
+        final me.leoko.advancedban.utils.Command command = me.leoko.advancedban.utils.Command.getByName(cmd.substring(1));
+        if (command != null && source instanceof Player) {
+            if (command.getPermission() == null || Universal.get().getMethods().hasPerms(source, command.getPermission())) {
```
A248, line 32: These can be collapsed.

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
```diff
@@ -0,0 +1,57 @@
...
+        public static final String PLUGIN_NAME = "AdvancedBan";
+    }
+    private static VelocityMain instance;
+    public final ProxyServer server;
+    public final Logger logger;
```
A248, line 31: I would discourage using public fields. Make a couple getters instead.

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,368 @@
...
+
+    @Override
+    public void setCommandExecutor(String cmd, TabCompleter tabCompleter) {
+        getPlugin().server.getCommandManager().register(new CommandReceiverVelocity(cmd), cmd);
+    }
+
+    @Override
+    public void sendMessage(Object player, String msg) {
+        if (player instanceof Player) {
+            ((Player) player).sendMessage(TextComponent.of(msg));
```
A248, line 139: Likewise with #hasPermission, this is an unnecessary cast. Also, are you sure `TextComponent.of` is the correct method? I thought the LegacyComponentSerializer was supposed to be used, although it's possible TextComponent.of will work for now.

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,368 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonNull;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.velocitypowered.api.command.CommandSource;
+import com.velocitypowered.api.proxy.Player;
+import de.leonhard.storage.LightningBuilder;
+import de.leonhard.storage.Yaml;
+import de.leonhard.storage.internal.settings.ConfigSettings;
+import jdk.nashorn.internal.parser.JSONParser;
```
A248, line 12: Why is this here?

core/src/main/resources/MySQL.yml
fede1132: As I already said before.. the library I used require an InputStream to inizialize the file, but of you look at the First lines for VelocityMethods mysql file Will be loader ONLY if the file exists

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
fede1132: Oh ok, my idea was different at begininng so I forgot to update this

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
fede1132: As I said the idea at begininng was different.. so I forgot to update it

DevLeoko:
Velocity Support has been added though #434 (and this PR looks inactive) so I'll close this PR. Thanks though for your effort.

PR #428, opened 2020-11-05
Refactor and consolidate configuration file loading

A248:
This started as a way to make support easier by making configuration loading fail-fast, but I realized there was a lot of duplicate code which could be reduced so did that as well.

This PR has no functional changes except by making the configuration loading fail-fast.

AdvancedBan used to load configurations using YamlConfiguration.load, but this static method is pretty poorly designed. Its fail-safe behaviour returns an empty (or partially complete?) YamlConfiguration instance. Since various parts of AdvancedBan assume configuration sections to exist, this causes NPEs to be thrown:

https://paste.hopefuls.de/bunicijuvu.bash
https://pastebin.com/cv40pGJw

With these changes, the user will be notified immediately of the problem, rather than having AdvancedBan continue on in a zombified state throwing intermittent NPEs.


DevLeoko (approved):
Looks good, thanks for contributing

PR #434, opened 2020-11-10
AdvancedBan Velocity 1.1.2

FawksX:
# Velocity 1.1.0

A previous Velocity Pull Request is already existing for AdvancedBan. However, I've decided to make a new one due to it being inactive for so long and many API changes (such as Adventures API changes) and make my own PR.

This will have numerous benefits including the use of [Sponge Configurate](https://github.com/SpongePowered/Configurate) usage.

This PR also implements a `ServerType` enum used for determining the platform which could be leveraged in certain applications. `isBungee` has also been deprecated in all settings and instead replaced with `isProxy`


A248:
A little background on slf4j first -
slf4j is a logging facade with multiple adapters for different logging frameworks. It's designed to be used where code does not necessarily know the underlying logging framework.

HikariCP, a connection pooling library which AdvancedBan uses, is a good example of a library which needs to perform logging. It would not make sense for HikariCP to force any of its dependents to use a particular logging framework, therefore it uses slf4j.

Paper, Waterfall, and Velocity being sane platforms for developing plugins, slf4j is already included on the classpath. It is also exposed in the API with convenience methods (though use of LoggerFactory directly is the idiomatic slf4j pattern, which is also all that matters in this case). Spigot and BungeeCord do not include slf4j, and AdvancedBan must unfortunately provide support for them, so it cannot rely on the presence of slf4j. To any users reading this, please use Paper, Waterfall, or Velocity, as it makes quality plugins far easier to develop.

The natural solution to a missing library is to shade and relocate it. This is what AdvancedBan does for slf4j. using slf4j-jdk14 as the adapter so that any logging framework capable of catching java.util.logging (jul) messages will handle them. This makes logging slightly less efficient on servers with slf4j already included, because there are several layers of translation (relocated slf4j -> relocated slf4j-jdk14 -> jul -> logging framework) rather than a simple abstraction and implementation (slf4j -> logging framework).

The relocation applied by the maven-shade-plugin means that any references to `org.slf4j` in the AdvancedBan codebase as well as dependencies such as HikariCP are rewritten to AdvancedBan's namespace. For Paper/Spigot and Waterfall/Bungee, this is fine, because those platforms expose jul plugin loggers.

For Velocity, there is no plugin.getLogger() or equivalent which returns a jul logger. There is only a slf4j logger, which is injected. This is optimal, but it ironically conflicts with a sub-optimal setup.

Options which can be taken to remedy this:
* Use java.util.logging on Velocity. plugin.getLogger() and similar constructs returning some kind of logger exist on all platforms, but they're really just convenience methods to provide loggers which prefix messages with the plugin name. However, it will not hurt too much to use `java.util.logging.Logger#getLogger` on Velocity, but it would need to be tested to ensure Velocity handles jul messages.
* Manipulate the multi-module structure to apply relocation only to slf4j and HikariCP which depends on it. The maven-shade-plugin would have to be added to the core module. There, slf4j and HikariCP would be shaded with relocations applied, and slf4j would be marked as a runtime dependency. The Velocity module would have no shading or relocation applied to it, except of course its inclusion into the bundle module output jar.
* Split the Velocity plugin jar. This would enable us greater flexibility as to managing dependencies for the Velocity plugin. It would fit well with some of the other necessities of providing Velocity support, such as the fact Velocity does not ship with any JDBC drivers. AdvancedBan will need the Connector-J or MariaDB-Connector, JDBC drivers for MySQL/MariaDB which are provided on Paper/Spigot and Waterfall/Bungee. Including these drivers on all platforms would cause conflicts due to the antiquated URL-driven system used by DriverManager (others reading this, prefer javax.sql.DataSource)

FawksX:
I saw in the previous PR you suggested splitting the Velocity Module - Which I feel is the best course of action if you agree with me again this time. It'd reduce the size of the plugin substantially.

If so, I can make another commit in the next little bit to split the modules and import the JDBC Driver, etc for Velocity. Let me know which solution you think is best.

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,371 @@
...
+    return ((ConfigurationNode)file).getNode(path).getInt();
+  }
+
+  @Override
+  public List<String> getStringList(Object file, String path) {
+    try {
+      return ((ConfigurationNode)file).getNode(path).getList(TypeToken.of(String.class));
+    } catch (ObjectMappingException e) {
+      e.printStackTrace();
+      return null;
```
A248, line 307: Rethrow as an unchecked exception and fail-fast rather than returning null and causing further NPEs down the line.

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
```diff
@@ -0,0 +1,48 @@
...
+import me.leoko.advancedban.manager.CommandManager;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CommandRecieverVelocity implements SimpleCommand {
+
+  @Inject
+  private ProxyServer server;
```
A248, line 18: You can't inject into objects not created through Guice.

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @MonotonicNonNull private final String cmd;
```
A248, line 20: Should not be MonotonicNonNull

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
```diff
@@ -0,0 +1,48 @@
...
+    if (invocation.arguments().length > 0) {
+      args[0] = (this.server.getPlayer(args[0]).isPresent() ? this.server.getPlayer(args[0]).get().getUsername() : args[0]);
+    }
+    CommandManager.get().onCommand(commandSource, cmd, args);
+  }
+
+  @Override
+  public List<String> suggest(Invocation invocation) {
+    final me.leoko.advancedban.utils.Command command = me.leoko.advancedban.utils.Command.getByName(cmd.substring(1));
+    if (command != null && invocation.source() instanceof Player) {
```
A248, line 41: Why would `command` be null? It wouldn't make sense for CommandReceiverVelocity to pickup commands not used by AdvancedBan

velocity/src/main/java/me/leoko/advancedban/velocity/listener/ConnectionListenerVelocity.java
```diff
@@ -0,0 +1,21 @@
...
+import net.kyori.text.TextComponent;
+
+public class ConnectionListenerVelocity {
+
+  @Subscribe
+  public void onLogin(LoginEvent event) {
+    UUIDManager.get().supplyInternUUID(event.getPlayer().getUsername(), event.getPlayer().getUniqueId());
+    String result = Universal.get().callConnection(event.getPlayer().getUsername(), event.getPlayer().getRemoteAddress().getAddress().getHostAddress());
+    if (result != null) {
+      event.setResult(ResultedEvent.ComponentResult.denied(TextComponent.of(result)));
```
A248, line 17: net.kyori.text.TextComponent#of doesn't seem like the right class or method. For one, you should use net.kyori.adventure. For two, I'm pretty sure you're supposed to use the legacy component serializer for this.

velocity/src/main/java/me/leoko/advancedban/velocity/listener/ConnectionListenerVelocity.java
```diff
@@ -0,0 +1,21 @@
...
+import me.leoko.advancedban.Universal;
+import me.leoko.advancedban.manager.UUIDManager;
+import net.kyori.text.TextComponent;
+
+public class ConnectionListenerVelocity {
+
+  @Subscribe
+  public void onLogin(LoginEvent event) {
+    UUIDManager.get().supplyInternUUID(event.getPlayer().getUsername(), event.getPlayer().getUniqueId());
+    String result = Universal.get().callConnection(event.getPlayer().getUsername(), event.getPlayer().getRemoteAddress().getAddress().getHostAddress());
```
A248, line 15: This can be considerably cleaned up if you extract event.getPlayer to a local variable

velocity/src/main/java/me/leoko/advancedban/velocity/listener/ChatListenerVelocity.java
```diff
@@ -0,0 +1,24 @@
...
+public class ChatListenerVelocity {
+
+  @Subscribe
+  public void onChat(PlayerChatEvent event) {
+    if (!event.getMessage().startsWith("/")) {
+      if (Universal.get().getMethods().callChat(event.getPlayer())) {
+        event.setResult(PlayerChatEvent.ChatResult.denied());
+      }
+    } else {
+      if (Universal.get().getMethods().callCMD(event.getPlayer(), event.getMessage())) {
```
A248, line 16: else + if -> else-if
Even though you followed the example of ChatListenerBungee, it's better to introduce cleaner code going forward.

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+
+public class CommandRecieverVelocity implements SimpleCommand {
+
+  @Inject
+  private ProxyServer server;
+
+  @MonotonicNonNull private final String cmd;
+  @MonotonicNonNull private String[] args;
+  @MonotonicNonNull private CommandSource commandSource;
```
A248, line 22: I don't know where you got the idea that these should be field variables shared across invocations of `execute` and `suggest`. It looks like they should be local to both methods.

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public List<String> suggest(Invocation invocation) {
+    final me.leoko.advancedban.utils.Command command = me.leoko.advancedban.utils.Command.getByName(cmd.substring(1));
+    if (command != null && invocation.source() instanceof Player) {
+      if (command.getPermission() == null || Universal.get().getMethods().hasPerms(commandSource, command.getPermission())) {
+        return command.getTabCompleter().onTabComplete(commandSource, args);
+      }
+    }
+    return new ArrayList<>();
```
A248, line 46: Collections.emptyList() is a minor improvement

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  @Override
+  public boolean isBungee() {
+    return false;
+  }
+
+  @Override
+  public String clearFormatting(String text) {
+    return text.replaceAll("-\\d{1}", "");
```
A248, line 114: This regex pattern doesn't seem to match color codes. Are you sure this is right?

velocity/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.2.3</version>
+                <configuration>
+                    <relocations>
```
A248, line 54: I'm pretty sure this maven-shade-plugin configuration doesn't need to be duplicated. If you need to, apply shading of HikariCP in AdvancedBan-Core and make sure platform modules don't reference it by setting the right dependency scope.

A248 (changes requested):
Yes, I think distributing a separate Velocity jar would be the best way to go. Even though AdvancedBan normally uses Connector-J for MySQL support, I recommend you use MariaDB-Connector on the Velocity plugin as it is generally better maintained than Connector-J, and still has MySQL compat.

I've submitted a review of the progress so far for now. There are some suggestions and some things which need to be changed for certain.

velocity/src/main/java/me/leoko/advancedban/velocity/listener/ChatListenerVelocity.java
FawksX: Copied this and CommandRecieverVelocity from the old PR to save some time, after reading it again I'll be re-writing the entire thing

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
```diff
@@ -0,0 +1,52 @@
...
+import me.leoko.advancedban.Universal;
+import me.leoko.advancedban.velocity.listener.ChatListenerVelocity;
+import me.leoko.advancedban.velocity.listener.ConnectionListenerVelocity;
+
+import java.nio.file.Path;
+
+@Plugin(
+    id = "advancedban",
+    name = "AdvancedBan",
+    version = "${project.version}"
```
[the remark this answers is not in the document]
FawksX: I believe this can be avoided by changing the version during compilation as Velocity generates a velocity-plugin.json - Do you know of a way to filter this file while package is being ran?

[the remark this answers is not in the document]
A248: The velocity-plugin.json is generated to target/classes/. It would not be a good idea to make that output folder a resource directory.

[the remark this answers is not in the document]
FawksX: yeah you're right - I'm not gonna worry about this too much at the minute as it's a pretty minor thing to fix

FawksX:
Just pushed a bunch of fixes and the majority of the changes A248 requested above. A few things I haven't touched at the minute:

- The pom.xml for Velocity is 100% not final, I've just thrown it together to allow for initial testing
- There's still a ton to do in terms of development to ensure that this is going to work correctly

This is mainly touching up the issues listed above from my sleep deprived coding sessions but also to ensure that messages are sent correctly from Sponge Configurate which I have since completely tested and have found works perfectly.

Will be continuing this later

core/src/main/java/me/leoko/advancedban/MethodInterface.java
```diff
@@ -68,8 +68,13 @@
      */
     void setupMetrics();
 
+    @Deprecated
     boolean isBungee();
 
+    boolean isProxy();
+
+    String getProxyType();
```
A248, line 76: This is a good idea to add getProxyType(), but I'm wondering if it would make sense to use `getServerType()` instead. Then, BukkitMethods would return "Spigot" and the logic in the /advancedban would be simplified.

Perhaps even better than that would be a `ServerType` enum so code which uses getProxyType/getServerType isn't stringly-typed. That would replace `Universal.get().getProxyType().equals("Bungeecord")` with `Universal.get().getServerType() == ServerType.BUNGEECORD`

velocity/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+        <dependency>
+            <groupId>net.luckperms</groupId>
+            <artifactId>api</artifactId>
+            <version>5.2</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>3.8.1</version>
```
A248, line 45: I'm not sure why you're using such an old version of Junit. AdvancedBan currently uses Junit5 which should cover all the capability we need for tests.

velocity/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mariadb.jdbc</groupId>
+            <artifactId>mariadb-java-client</artifactId>
+            <version>2.1.2</version>
```
A248, line 51: 2.7.0 is the latest version, released in September

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private final ProxyServer server;
+  private final Path dataDirectory;
+  private final Logger logger;
+
+  private static VelocityMain instance;
+
+  @Inject
+  public VelocityMain(ProxyServer server, @DataDirectory Path dataDirectory, Logger logger) {
+    this.instance = this;
```
A248, line 32: `instance` is not an instance field (ironically), so this assignment needs to reflect that

bundle/pom.xml
```diff
@@ -40,10 +40,10 @@
                     		<pattern>org.apache.commons</pattern>
                     		<shadedPattern>${project.groupId}.shaded.org.apache.commons</shadedPattern>
                     	</relocation>
-                    	<relocation>
-                    		<pattern>org.slf4j</pattern>
-                    		<shadedPattern>${project.groupId}.shaded.org.slf4j</shadedPattern>
-                    	</relocation>
+                        <relocation>
+                            <pattern>org.slf4j</pattern>
```
A248, line 44: At the moment this is unnecessary diff

A248 (changes requested):
Looking good, still some more work to do though

core/src/main/java/me/leoko/advancedban/MethodInterface.java
FawksX: Yeah - I was thinking about this when I wrote this but ran out of time at the time. I feel like it could be useful for this

velocity/pom.xml
FawksX: I've not actually written a finished pom.xml for the velocity module as I haven't even touched the JCDB connectors for storage

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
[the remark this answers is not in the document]
FawksX: Fixed in next commit

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,365 @@
...
+      return json.get(keys[keys.length - 1]).toString().replaceAll("\"", "");
+
+    } catch (Exception exc) {
+      return null;
+    }
+  }
+
+  @Override
+  public String getVersion() {
+    return server.getVersion().toString();
```
A248, line 90: Using ProxyVersion#toString is not right. getVersion needs to be implemented to return the version of AdvancedBan.

A248 (changes requested):
Some more things I noticed

FawksX:
It compiles and runs!

I believe everything except database management is now in place. I've ran out of time today to fix the issues with relocations and the Hikari Driver Instances but by the looks of things it should be a pretty easy fix (unsure until I actually look into it)

core/src/main/java/me/leoko/advancedban/ServerType.java
```diff
@@ -0,0 +1,20 @@
...
+  SPIGOT("Spigot"),
+  TEST("Test");
+
+  private String serverType;
+
+  private ServerType(String serverType) {
+    this.serverType = serverType;
+  }
+
+  public String getString() {
```
A248, line 16: toString() would make more sense here than getString

FawksX: Agreed, will change this on the next commit

FawksX:
It compiles and stores in both MySQL and HSQLDB with no issues, everything seems to load and messages can be edited with no issues!

@A248 - You have more experience with the codebase and the final issue which I can see with this is offline players. You cannot unban, ban etc if they're offline - Could you help propose a fix?

core/src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
```diff
@@ -18,13 +19,21 @@ public DynamicDataSource(boolean preferMySQL) {
...
+
             config.setJdbcUrl("jdbc:mysql://" + ip + ":" + port + "/" + dbName + "?"+properties);
             config.setUsername(usrName);
             config.setPassword(password);
         } else {
             config.setJdbcUrl("jdbc:hsqldb:file:" + mi.getDataFolder().getPath() + "/data/storage;hsqldb.lock_file=false");
             config.setUsername("SA");
             config.setPassword("");
+
+            if(Universal.get().getServerType() == ServerType.VELOCITY) {
```
A248, line 34:
```suggestion
            
```

core/src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
```diff
@@ -18,13 +19,21 @@ public DynamicDataSource(boolean preferMySQL) {
...
+                config.setDriverClassName("org.hsqldb.jdbc.JDBCDriver");
```
A248, line 35: Good catch with setting the driver class name. For HSQLDB, setting the driver class name should actually be unconditional and applied for all platforms. (For the MariaDB driver, the code you have is correct because AdvancedBan uses the inbuilt driver on platforms besides Velocity)

A248 (changes requested):
For the commands, I believe the same limitations are present on BungeeCord. The UUIDManager will be used to lookup the target player's UUID depending on the FetcherMode, which is by default detected based on the environment. If you are testing your PR in offline mode, mi.isOnlineMode() will be false and no UUIDs will be looked up.

I realized you are right with regards to setting the JDBC driver class name - it ensures AdvancedBan uses its own relocated driver, even when another plugin shades a different version of the same driver.

There are a couple other comments that still need to be addressed, but otherwise this PR is close to ready. We can then get one more reviewer (probably Leoko)

core/src/main/java/me/leoko/advancedban/utils/DynamicDataSource.java
FawksX: > For the commands, I believe the same limitations are present on BungeeCord. The UUIDManager will be used to lookup the target player's UUID depending on the FetcherMode, which is by default detected based on the environment. If you are testing your PR in offline mode, mi.isOnlineMode() will be false and no UUIDs will be looked up.
> 
> I realized you are right with regards to setting the JDBC driver class name - it ensures AdvancedBan uses its own relocated driver, even when another plugin shades a different version of the same driver.
> 
> There are a couple other comments that still need to be addressed, but otherwise this PR is close to ready. We can then get one more reviewer (probably Leoko)

This is the issue - I'm testing in online mode. I'll have a look into the UUIDManager and see if I can figure it out.

Realised I missed a few of the requested changes above, so I'll get them done for the next commit.

FawksX: Issue is Velocity has no built-in method of fetching a UUID - Will write something using the Mojang API in the next little bit in order to fetch a uuid.

A248: That's exactly what UUIDManager does - makes a request to the Mojang API or another web service as configured

FawksX: Well, need to diagnose why it's not working then, xD

FawksX:
Finished all previously requested changes.

As spoken above, the only issue now is that it is not storing UUIDs at all and to be able to unban just names it requires a proxy reboot - It's a pretty major issue which is preventing this from being finished.

At the minute, my main suspicion is VelocityMethods#getInternUUID (Object-based) as in the files it is returning "none" instead of an offline/online users uuid and I can't find any other code references which uses "none"

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
```diff
@@ -0,0 +1,45 @@
...
+  public List<String> suggest(Invocation invocation) {
+    String[] args = invocation.arguments();
+    CommandSource source = invocation.source();
+    try {
+      final me.leoko.advancedban.utils.Command command = me.leoko.advancedban.utils.Command.getByName(cmd.substring(1));
+      if (command.getPermission() == null || Universal.get().getMethods().hasPerms(source, command.getPermission())) {
+        return command.getTabCompleter().onTabComplete(source, args);
+      }
+      return Collections.emptyList();
+    } catch (NullPointerException e){
```
A248, line 41: Why are you catching NullPointerException? We need less swallowed exceptions in AdvancedBan, not more. Use a null-check instead.

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
```diff
@@ -0,0 +1,62 @@
...
+
+  private final ProxyServer server;
+  private final Path dataDirectory;
+  private final Logger logger;
+
+  private static VelocityMain velocityMain;
+
+  @Inject
+  public VelocityMain(ProxyServer server, @DataDirectory Path dataDirectory, Logger logger) {
+    this.velocityMain = this;
```
A248, line 36:
```suggestion
    velocityMain = this;
```
velocityMain is not an instance variable

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
```diff
@@ -0,0 +1,62 @@
...
+import me.leoko.advancedban.velocity.listener.ChatListenerVelocity;
+import me.leoko.advancedban.velocity.listener.ConnectionListenerVelocity;
+import org.slf4j.Logger;
+
+import java.nio.file.Path;
+
+@Plugin(
+    id = "advancedban",
+    name = "AdvancedBan",
+    version = "2.3.0",
```
A248, line 21: @ version needs to be brought back here, same goes for VelocityMethods#getVersion

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,375 @@
...
+  }
+
+  @Override
+  public String getVersion() {
+    return "2.3.0";
+  }
+
+  @Override
+  public String[] getKeys(Object file, String path) {
+    return new String[0];
```
A248, line 101: Needs to be implemented

A248 (changes requested):

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMain.java
FawksX: Oops, must've overriden this when I pulled it

A248:
Are you able to figure out what makes the current situation on BungeeCord different? That may give you more clues which could help in debugging.

Also, did you see my comment from a week ago about VelocityMethods#getStringList?

FawksX:
I've just noticed I never addressed the VelocityMethods#getStringList change - it was collapsed in this commit and I'll get to it next time I work on this.

Honestly - Not really sure what makes it different from bungeecord as they're both using the Mojang API to fetch the UUID. It was also an issue present in #399

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
FawksX: Unless I'm mistaken - I cannot rethrow an ObjectMappingException as an UncheckedIOException? Unless of course you mean a different unchecked exception

FawksX:
Apologies for no updates the last couple of days, have had a large influx of school work and sickness, plan to get this done in the next week or so.

For an update, still looking into the issue with the UUIDManager - It still doesn't have any glaring differences but hopefully once I have more time the small break might provide some clarity

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
FawksX: Eh, this Try-catch and a null-check isn't even needed. For some reason the CleanTabCompleter is not returning any players despite my testing showing it's able to find them. Will be looking into this more.

FawksX:
Nearing Completion!

With some further testing UUIDManager is working now- I don't know if my Velocity instance was messed up or what, but it seems to be working now. Won't be touching that anymore.

One single issue is still apparent which is what the null-pointer was being swallowed for - which is the TabCompletion not adding players. I've ran out of time to look into this, but once this is complete all functionality is restored to the plugin on a Velocity proxy from my testing.

Going to be marking this as a ready for review as there's only one issue left to get reviewed once I've fixed this last issue

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandRecieverVelocity.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.velocitypowered.api.command.CommandSource;
+import com.velocitypowered.api.command.SimpleCommand;
+import com.velocitypowered.api.proxy.ProxyServer;
+import me.leoko.advancedban.Universal;
+import me.leoko.advancedban.manager.CommandManager;
+import java.util.Collections;
+import java.util.List;
+
+public class CommandRecieverVelocity implements SimpleCommand {
```
A248, line 11:
```suggestion
public class CommandReceiverVelocity implements SimpleCommand {
```
Just noticed this was misspelled

core/src/main/java/me/leoko/advancedban/ServerType.java
```diff
@@ -0,0 +1,20 @@
...
+  SPIGOT("Spigot"),
+  TEST("Test");
+
+  private String serverType;
+
+  private ServerType(String serverType) {
+    this.serverType = serverType;
+  }
+
+  public String toString() {
```
A248, line 16: Needs `@Override`

velocity/src/main/java/me/leoko/advancedban/velocity/event/RevokePunishmentEvent.java
```diff
@@ -0,0 +1,32 @@
+package me.leoko.advancedban.velocity.event;
+
+import me.leoko.advancedban.utils.Punishment;
+
+public class RevokePunishmentEvent {
+  private final Punishment punishment;
+  private final boolean massClear;
+
+  public RevokePunishmentEvent(Punishment punishment, boolean massClear) {
+    this.punishment = punishment;
```
A248, line 10: It would be a good idea to validate that `Punishment` is non-null, e.g. using Objects.requireNonNull.

Same goes for PunishmentEvent

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
A248: RuntimeException or any subclass which makes sense is fine:
```java
try {
  // configurate which throws checked exceptions
} catch (ObjectMappingException ex) {
  throw new RuntimeException(ex);
}
```

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,377 @@
...
+
+  @Override
+  public boolean isOnline(String name) {
+    return server.getPlayer(name).isPresent();
+  }
+
+  @Override
+  public Player getPlayer(String name) {
+    if(server.getPlayer(name).isPresent()) {
+      return server.getPlayer(name).get();
```
A248, line 202:
```suggestion
      return server.getPlayer(name).orElse(null;
```
Would make the code cleaner and also avoid a double map lookup

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  @Override
+  public String getVersion() {
+    return "@version";
+  }
+
+  @Override
+  public String[] getKeys(Object file, String path) {
+    return ((ConfigurationNode)file).getNode(path.split("\\.")).getChildrenMap().keySet().stream().map(Object::toString).toArray(String[]::new);
```
A248, line 102: It might be worth adding a comment that relying on Object.toString() is safe here - I had to go and check the Configurate docs to be sure

A248 (commented):
A couple small things I found

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  @Override
+  public String getName(String uuid) {
+      return server.getPlayer(UUID.fromString(uuid)).get().getUsername();
+  }
+
+  @Override
+  public String getIP(Object player) {
+    return ((Player)player).getRemoteAddress().getAddress().getHostAddress();
```
A248, line 169:
```suggestion
    return ((Player)player).getRemoteAddress().getHostName();
```
For conformity with BukkitMethods and BungeeMethods

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,377 @@
...
+
+  @Override
+  public boolean isOnline(String name) {
+    return server.getPlayer(name).isPresent();
+  }
+
+  @Override
+  public Player getPlayer(String name) {
+    if(server.getPlayer(name).isPresent()) {
+      return server.getPlayer(name).orElse(null);
```
A248, line 203: This entire method implementation can simply be `return server.getPlayer(name).orElse(null);`

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  @Override
+  public void runSync(Runnable rn) {
+    rn.run();
+  }
+
+  @Override
+  public void executeCommand(String cmd) {
+    server.getCommandManager().executeImmediatelyAsync(server.getConsoleCommandSource(), cmd);
```
A248, line 240: Two things here:
1. This should probably use `executeAsync` rather than `executeImmediatelyAsync`, as I don't see a reason to skip the CommandExecuteEvent.

2. Per CompletableFuture best practices, an exception handler should be attached to the returned CompletableFuture, since this future is neither chained nor returned nor awaited:
```java
methodReturningFuture().exceptionally((ex) -> {
  ex.printStackTrace();
  return null;
});
```
executeImmediatelyAsync and executeAsync both return CompletableFuture

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,377 @@
...
+        sendMessage(player, str);
+      }
+      return true;
+    }
+    return false;
+  }
+
+  @Override
+  public String parseJSON(InputStreamReader json, String key) {
+    JsonElement element = new JsonParser().parse(json);
```
A248, line 270: JsonParser constructor is deprecated

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,377 @@
...
+
+  @Override
+  public String getVersion() {
+    return "@version";
+  }
+
+  // Object.toString() is considered safe in this scenario (as per Configurate Docs)
+  @Override
+  public String[] getKeys(Object file, String path) {
+    return ((ConfigurationNode)file).getNode(path.split("\\.")).getChildrenMap().keySet().stream().map(Object::toString).toArray(String[]::new);
```
A248, line 103: Getting a node from a path is copied a lot. You could make a private method to reduce code duplication:
```java
private ConfigurationNode getConfigNode(Object file, String path) {
  return ((ConfigurationNode) file).getNode((Object[]) path.split("\\."));
}
```

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandReceiverVelocity.java
```diff
@@ -0,0 +1,41 @@
+package me.leoko.advancedban.velocity.command;
```
A248, line 1: In accordance with the layout of other AdvancedBan environments (Bukkit and Bungee), perhaps you could move this to the `listener` package

A248 (commented):
I also noticed when building AdvancedBan with this PR that you are shading slf4j. Velocity has slf4j included, so it should not be shaded in AdvancedBan-velocity.

velocity/src/main/java/me/leoko/advancedban/velocity/command/CommandReceiverVelocity.java
FawksX: I've refactored this but I do believe that this should be moved to its own package as it's not a Listener

FawksX:
> I also noticed when building AdvancedBan with this PR that you are shading slf4j. Velocity has slf4j included, so it should not be shaded in AdvancedBan-velocity.

This is currently Included in the core module - Is it preferred it's included in both Bungee and bukkit modules or just Bundle?

A248:
Slf4j is a transitive dependency of the core module, and correctly so. However, the existence of a dependency does not necessitate it be shaded. There are a few ways you can prevent a dependency from being shaded when using the shade plugin.

In the AdvancedBan-velocity module, re-declare the slf4j-api dependency in the provided scope. Then move the slf4j-jdk14 dependency from the core module to the bundle module, keeping it as runtime scope. This will prevent slf4j-api and slf4j-jdk14 from being shaded on Velocity, while retaining them as dependencies in the right places.

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -267,76 +266,76 @@ public boolean callCMD(Object player, String cmd) {
 
   @Override
   public String parseJSON(InputStreamReader json, String key) {
-    JsonElement element = new JsonParser().parse(json);
-    if (element instanceof JsonNull) {
+    JsonObject element = (JsonObject) JsonParser.parseReader(json);
+    if (element.isJsonNull()) {
```
A248, line 270: One thing or another is wrong here. Either:
* `element` is always a JsonObject, in which case it will never be a JsonNull.
* `element` can be a JsonNull, in which case the cast to JsonObject is premature, and will cause ClassCastException.

The isJsonXXX methods perform simple instanceof checks - arguably a poor decision from the Gson team. I'd recommend using `instanceof JsonNull` rather than `isJsonNull()`, for readability reasons.

FawksX: the main reason behind this is because IntelliJ complains and says the two types cannot be casted. Instead, I've made a try-catch for getAsJsonObject() which in turn throws IllegalStateException where I can then return null.

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -266,21 +266,25 @@ public boolean callCMD(Object player, String cmd) {
...
   }
 
   @Override
   public String parseJSON(String json, String key) {
-    JsonObject element = (JsonObject) JsonParser.parseString(json);
-    if (element.isJsonNull()) {
+    JsonObject element;
+    try {
+      element = (JsonObject) JsonParser.parseString(json);
+    } catch (IllegalStateException e) {
```
A248, line 284: Did you mean to use getAsJsonObject instead of a cast here? That would make sense as to why you're catching IllegalStateException

core/pom.xml
```diff
@@ -51,10 +51,10 @@
 
         <dependency>
             <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-jdk14</artifactId>
+            <artifactId>slf4j-api</artifactId>
             <!-- Should be the same version as HikariCP uses -->
             <version>1.7.25</version>
-            <scope>runtime</scope>
+            <scope>provided</scope>
```
A248, line 57: Do make sure that slf4j is still shaded on bungee and bukkit

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
FawksX: Yes - my bad, I forgot to change it.

core/pom.xml
FawksX: It is shaded in the bundle module

core/src/main/java/me/leoko/advancedban/Universal.java
```diff
@@ -142,6 +142,14 @@ public boolean isBungee() {
         return mi.isBungee();
```
A248, line 142: Since MethodInterface#isBungee is deprecated, Universal#isBungee should also be so

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  @Override
+  public Player[] getOnlinePlayers() {
+    return server.getAllPlayers().toArray(new Player[]{});
+  }
+
+  @Override
+  public void scheduleAsyncRep(Runnable rn, long l1, long l2) {
+    server.getScheduler().buildTask(getPlugin(), rn).delay(l1*50, TimeUnit.MILLISECONDS).repeat(l2*50, TimeUnit.MILLISECONDS);
```
A248, line 216: Missing call to .schedule()
```suggestion
    server.getScheduler().buildTask(getPlugin(), rn).delay(l1*50, TimeUnit.MILLISECONDS).repeat(l2*50, TimeUnit.MILLISECONDS).schedule();
```

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  @Override
+  public String getFileName(Object file) {
+    return "[Only Available in the Bukkit Version!]";
+  }
+
+  @Override
+  public void callPunishmentEvent(Punishment punishment) {
+    server.getEventManager().fire(new PunishmentEvent(punishment));
```
A248, line 352: `fire` returns a CompletableFuture, so use fireAndForget instead
```suggestion
    server.getEventManager().fireAndForget(new PunishmentEvent(punishment));
```

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  @Override
+  public void callRevokePunishmentEvent(Punishment punishment, boolean massClear) {
+    server.getEventManager().fire(new RevokePunishmentEvent(punishment, massClear));
```
A248, line 357:
```suggestion
    server.getEventManager().fireAndForget(new RevokePunishmentEvent(punishment, massClear));
```

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -0,0 +1,384 @@
...
+
+  @Override
+  public boolean isOnlineMode() {
+    return server.getConfiguration().isOnlineMode();
+  }
+
+  @Override
+  public void notify(String perm, List<String> notification) {
+    server.getAllPlayers().forEach(player -> {
+      if(player.hasPermission(perm)) notification.forEach(str -> sendMessage(player, str));});
```
A248, line 368: I would recommend cleaning this up slightly:
```suggestion
    server.getAllPlayers().forEach(player -> {
        if (player.hasPermission(perm)) {
            notification.forEach(str -> sendMessage(player, str));
        }
    });
```

velocity/src/main/java/me/leoko/advancedban/velocity/listener/ChatListenerVelocity.java
```diff
@@ -0,0 +1,22 @@
...
+
+  @Subscribe
+  public void onChat(PlayerChatEvent event) {
+    if (!event.getMessage().startsWith("/")) {
+      if (Universal.get().getMethods().callChat(event.getPlayer())) {
+        event.setResult(PlayerChatEvent.ChatResult.denied());
+      }
+    } else if (Universal.get().getMethods().callCMD(event.getPlayer(), event.getMessage())) {
+        event.setResult(PlayerChatEvent.ChatResult.denied());
+      }
```
A248, line 17:
```suggestion
    }
```

velocity/src/main/java/me/leoko/advancedban/velocity/listener/ChatListenerVelocity.java
```diff
@@ -0,0 +1,22 @@
...
+    }
```
A248, line 18:
```suggestion
  }
```

A248 (commented):
I found a few more things

A248:
Just noticed 2 issues while building:

1. slf4j is being shaded into AdvancedBan-Velocity:
```
[INFO] --- maven-shade-plugin:3.2.3:shade (default) @ AdvancedBan-Velocity ---
[INFO] Including org.slf4j:slf4j-api:jar:1.7.30 in the shaded jar.
[INFO] Including me.leoko.advancedban:AdvancedBan-Core:jar:2.3.0 in the shaded jar.
[INFO] Including commons-io:commons-io:jar:2.5 in the shaded jar.
[INFO] Including org.apache.commons:commons-lang3:jar:3.8.1 in the shaded jar.
[INFO] Including org.hsqldb:hsqldb:jar:2.4.0 in the shaded jar.
[INFO] Including com.zaxxer:HikariCP:jar:3.4.5 in the shaded jar.
[INFO] Including org.mariadb.jdbc:mariadb-java-client:jar:2.7.0 in the shaded jar.
```
2. `@version` is replaced in the source files as well as the compiled files. Therefore, VelocityMain and VelocityMethods generate new diff every time you build, which is easy to accidentally commit.

It should be that these variables are replaced at compile-time (in the resulting jar), but not in the source code (the VelocityMain and VelocityMethods source files).

A248 (approved):
Code looks good to me

DevLeoko:
Thanks a bunch to both @FawksX for the PR and @A248 for overseeing it. This will open up AdvancedBan to a lot more users and has been requested quite often in the past few months. Geat work!

Issue #437, opened 2020-11-12
Remove Class.forName in DynamicDataSource


A248:
This is an antiquated approach to ensure a JDBC driver is loaded, which somehow stubbornly persists to this day. Since Java 6, the ServiceLoader mechanism is used to load java.sql.Driver implementations. The JDBC url is therefore sufficient to load Drivers.

Only very, very old JDBC drivers which are not JDBC 4.0-compliant will not have a ServiceLoader file. Fortunately, Connector-J introduced support in [5.1.3](https://dev.mysql.com/doc/relnotes/connector-j/5.1/en/news-5-1-3.html), way back in 2007.

This means the Class.forName call can be safely removed from DynamicDataSource

Issue #433, opened 2020-11-09
utf8/UTF-8 characterEncoding Connection URL property


A248:
The documentation of MySQL Connector-J suggests AdvancedBan's default `characterEncoding=utf8` is incorrect. It probably should be `UTF-8`, the "Java-style" name mentioned here: https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-charsets.html . It is the same for 5.1

> Use Java-style names when specifying character encodings

That implies utf8 is not the right value. Indeed, this seems to be confirmed by StackOverflow: https://stackoverflow.com/a/20423184

PR #438, opened 2020-11-12
Solve Issues 433 and 437: characterEncoding=UTF-8 and remove Class.forName on driver class

A248:
References: #433 and #437


DevLeoko (approved):
Thanks for contributing

PR #442, opened 2020-12-02
Better exception handling and debugging

A248:
LogManager and the practice of shunting away errors to AdvancedBan/logs/latest.log has made support harder with no noticeable benefit to users. If anything, it is more confusing when users are told they have another latest.log.

This PR removes LogManager and implements better exception handling with regards to SQLExceptions, including fail-fast behaviour and usage of the server console as it is intended.


DevLeoko (approved):
Looks good, thanks for contributing 👍

PR #449, opened 2020-12-30
Compatibility with JPremium and AuthMe + FastLogin for uuids using LuckPerms (Fixed #446)

### LINKED ISSUE #446: Feature: Use Luckperms for UUID-fetcher intern mode
 2.3.0
 
 Bungee
 
Latest FlameCord

---

Use Luckperms to find the UUID of an user when running in offline mode with Intern UUID fetcher. This is a better alternative than just using cache ^^

CustomEntity:
That Pull Request allows to use LuckPerms to get UUIDs when bungee is in offline mode and uuid fetcher bungee side is intern.

### DIFF
--- a/bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
+++ b/bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
@@ -17,6 +17,9 @@ import me.leoko.advancedban.manager.UUIDManager;
 import me.leoko.advancedban.utils.Permissionable;
 import me.leoko.advancedban.utils.Punishment;
 import me.leoko.advancedban.utils.tabcompletion.TabCompleter;
+import net.luckperms.api.LuckPerms;
+import net.luckperms.api.LuckPermsProvider;
+import net.luckperms.api.model.user.User;
 import net.md_5.bungee.api.ChatColor;
 import net.md_5.bungee.api.CommandSender;
 import net.md_5.bungee.api.ProxyServer;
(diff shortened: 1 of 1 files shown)


bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
```diff
@@ -277,6 +280,12 @@ public String getInternUUID(Object player) {
 
     @Override
     public String getInternUUID(String player) {
+        if (luckPermsSupport) {
+            UUID uuid = LuckPermsProvider.get().getUserManager().lookupUniqueId(player).join();
```
A248, line 284: Are you sure this is guaranteed to not throw NoClassDefFoundError when LuckPerms is not present? I have found that sometimes classloading still occurs despite using a branch.

A248 (commented):
There are 2 getInternUUID methods in BungeeMethods, but you only changed one of them. The other one should be modified similarly.

bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
DevLeoko: Yes, this will probably cause errors for users that don't use LuckPerms. There should be no import statements of the LuckPerms API in java files that are also in use without LuckPerms. 

We have avoided this issue for permissions checks with the `LuckPermsOfflineUser` class.

DevLeoko (changes requested):
See comments above for required changes

PR #450, opened 2021-01-08
Add setDriverClassName for hsqldb

A248:
Fixes a bug introduced in #438 , which causes a "No suitable driver" exception.

https://pastebin.com/q6Li9JNF

The fix is very simple.


DevLeoko (approved):

DevLeoko:
Thanks for the fix

PR #454, opened 2021-01-25
[WIP] Sponge API 7 Support

FawksX:
Sponge API 7 Support

This is **in progress** - It does compile, but doesn't start successfully. More to follow

This has been built alongside Sponge API 7.1 - mainly because thats the version I use. I have shaded in Configurate 3.7.1 so it is consistent across all versions. API 8s version will also probably use this version of configurate (unless there's a super good reason to use 4.0.0 for this)

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -56,5 +56,12 @@
             <!-- Usually provided by the server implementation -->
             <scope>provided</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.spongepowered</groupId>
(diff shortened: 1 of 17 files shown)


FawksX:
Functionality is there and banning/unbanning works. However, Layouts are broken (unsure as to why) - however, I've ran out of time for now.

If you are using Nucleus, make sure to disable their banning system under `/config/nucleus/config.conf`

RawrUniversal:
Any update on this? It would be nice to have this as a sponge plugin.

PR #469, opened 2021-02-25
Ensure the ChatListenerVelocity recieves the event early

STG-Allen:
This PR aims to make sure that the ChatListener on Velocity is called before other plugins, allowing them to easily check if the event is cancelled.


DevLeoko (approved):
Thanks for making AdvancedBan more compatible

PR #470, opened 2021-02-25
Register commands with permissions on Bungeecord

### LINKED ISSUE #440: CommandCompletion shows all commands despite missing permission
2.3.0 
 
Bungeecord 
 
Waterfall-Bootstrap:1.16-R0.4-SNAPSHOT:af18705:381

If a player without ban, mute etc. permission enters a "/" he will be shown all advancedban commands although he has no permission for it.

_Any error/log post it through [pastebin.com](http://pastebin.com) and link it here. (Also include /plugins/AdvancedBan/logs/latest.log)_
No

It's not bad but don't like the fact that there are commands for which you have no authorization.
Would be cool if you could change this on occasion :)

ArnaudLier:
This allows for not showing commands that the user can't execute in the autocompletion. Fixes #440

### DIFF
--- a/bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
+++ b/bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
@@ -14,6 +14,7 @@ import me.leoko.advancedban.bungee.utils.LuckPermsOfflineUser;
 import me.leoko.advancedban.manager.DatabaseManager;
 import me.leoko.advancedban.manager.PunishmentManager;
 import me.leoko.advancedban.manager.UUIDManager;
+import me.leoko.advancedban.utils.Command;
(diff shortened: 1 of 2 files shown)


DevLeoko (changes requested):
The `setCommandExecutor(...)` method is called from the universal class at which point there is a reference to the AdvancedBan Command object. Which is lost because only the command name and TabCompleter are passed to the method. Later retrieving the Command again based on the name is a little "hacky" as we already had that reference above the method call.

The proper way to go here would probably be to change the method signature of `setCommandExecutor` and also pass the permission required. This way you could also fix this issue for bukkit in the same way.

quiquelhappy:
a fix/merge for this would be appreciated

ArnaudLier:
Superseded by #516

PR #471, opened 2021-02-25
Add CloudNet-CloudPerms offline exempt perms support

ZeProf2Code:
Everything's in title.

### DIFF
--- a/bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
+++ b/bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
@@ -10,6 +10,7 @@ import me.leoko.advancedban.Universal;
 import me.leoko.advancedban.bungee.event.PunishmentEvent;
 import me.leoko.advancedban.bungee.event.RevokePunishmentEvent;
 import me.leoko.advancedban.bungee.listener.CommandReceiverBungee;
+import me.leoko.advancedban.bungee.utils.CloudNetCloudPermsOfflineUser;
 import me.leoko.advancedban.bungee.utils.LuckPermsOfflineUser;
 import me.leoko.advancedban.manager.DatabaseManager;
 import me.leoko.advancedban.manager.PunishmentManager;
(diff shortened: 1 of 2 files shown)


bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
```diff
@@ -187,7 +194,11 @@ public boolean hasPerms(Object player, String perms) {
 
     @Override
     public Permissionable getOfflinePermissionPlayer(String name) {
-        if(luckPermsSupport) return new LuckPermsOfflineUser(name);
+        if (permissionable != null) {
+            try {
+                return permissionable.getConstructor(String.class).newInstance(name);
```
DevLeoko, line 199: The java compiler will probably not be able to detect whether the class actually has this constructor. Using `Function<String, Permissionable> perissionableGenerator` might be a way of keeping it modular while allowing for proper code analysis by the compiler.

DevLeoko (changes requested):

bungee/src/main/java/me/leoko/advancedban/bungee/BungeeMethods.java
ZeProf2Code: Is this correct? If not, I'm sorry, I'm a beginner in Java development.

DevLeoko: Yes, this is perfect

bungee/src/main/java/me/leoko/advancedban/bungee/utils/CloudNetCloudPermsOfflineUser.java
```diff
@@ -0,0 +1,20 @@
...
+import de.dytanic.cloudnet.driver.CloudNetDriver;
+import de.dytanic.cloudnet.driver.permission.IPermissionUser;
+
+import java.util.UUID;
+
+public class CloudNetCloudPermsOfflineUser implements Permissionable {
+    private IPermissionUser permissionUser;
+
+    public CloudNetCloudPermsOfflineUser(String name) {
+        this.permissionUser = CloudNetDriver.getInstance().getPermissionManagement().getUsers(name).get(0);
```
DevLeoko, line 13: An ArrayList throws an IndexOutOfBounds exception if you try to access an invalid index (if there is no player with that name) instead of returning null which is what you would need for the check below to work properly.

DevLeoko (changes requested):
Didn't see that before, sorry ^^

Btw, did you test this?

bungee/src/main/java/me/leoko/advancedban/bungee/utils/CloudNetCloudPermsOfflineUser.java
```diff
@@ -0,0 +1,27 @@
...
+    private IPermissionUser permissionUser;
+
+    public CloudNetCloudPermsOfflineUser(String name) {
+    	final List <IPermissionUser> users = CloudNetDriver.getInstance().getPermissionManagement().getUsers(name);
+    	
+    	if (!users.isEmpty()) {
+        	this.permissionUser = users.get(0);
+    	} else {
+    		this.permissionUser = null;
+    	}
```
DevLeoko, line 20: The spacing is broken again.

Also no need for the else statement. Object attributes are null by default

DevLeoko (changes requested):
The spacing is now broken again

bungee/src/main/java/me/leoko/advancedban/bungee/utils/CloudNetCloudPermsOfflineUser.java
ZeProf2Code: oh yeah I thought it was final, mb

DevLeoko (approved):
Thank you very much for contributing to AdvancedBan 👍

DevLeoko:
Saw this too late now but PRs should be directed into the development branch. But no harm done, we have a release coming up soon anyway.

PR #484, opened 2021-04-07
Use compiled regex patterns anywhere is possible

iGabyTM:


core/src/main/java/me/leoko/advancedban/utils/Command.java
```diff
@@ -542,7 +543,13 @@ else if(args.length == 2 && !args[0].matches("\\d+"))
...
     Command(String permission, String regex, TabCompleter tabCompleter, Consumer<CommandInput> commandHandler,
             String usagePath, String... names) {
-        this(permission, (args) -> String.join(" ", args).matches(regex), tabCompleter, commandHandler, usagePath, names);
+        final Pattern pattern = Pattern.compile(regex);
+        this.permission = permission;
+        this.syntaxValidator = (args) -> pattern.matcher(String.join(" ", args)).matches();
+        this.tabCompleter = tabCompleter;
+        this.commandHandler = commandHandler;
+        this.usagePath = usagePath;
+        this.names = names;
```
DevLeoko, line 552: I think it's better to call the other constructor here (as it was before) to prevent code duplication

core/src/main/java/me/leoko/advancedban/utils/Regex.java
```diff
@@ -0,0 +1,147 @@
+package me.leoko.advancedban.utils;
+
+import java.util.regex.Pattern;
+
+/**
+ * @author <a href="https://github.com/iGabyTM">GabyTM</a>
+ */
+public enum Regex {
+
+    ZERO_OR_N_DIGITS("[0-9]*"),
```
DevLeoko, line 10: From what I see, this regex can be replaced with the `DIGITS` regex, as the case of no digits is irrelevant in the places where it is used

core/src/main/java/me/leoko/advancedban/utils/Regex.java
```diff
@@ -0,0 +1,147 @@
...
+    DIGITS("\\d+"),
+    DIGITS_OR_X("\\d+|X"),
+    IP("^(?:[0-9]{1,3}\\.){3}[0-9]{1,3}$"),
+    REASON_LAYOUT("[~@].+"),
+    TIME_LAYOUT("#.*"),
+    TWO_DIGITS("[1-9][0-9]*");
```
DevLeoko, line 16: This is actually a regex for digits >= 1 (not zero) and not two digits.

DevLeoko (changes requested):
Other than that everything looks good

bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
```diff
@@ -238,13 +239,13 @@ public String getIP(Object player) {
...
     public String getInternUUID(Object player) {
-        return player instanceof OfflinePlayer ? ((OfflinePlayer) player).getUniqueId().toString().replaceAll("-", "") : "none";
+        return player instanceof OfflinePlayer ? Regex.Replace.DASH.remove(((OfflinePlayer) player).getUniqueId().toString()) : "none";
     }
 
     @SuppressWarnings("deprecation")
     @Override
     public String getInternUUID(String player) {
-        return Bukkit.getOfflinePlayer(player).getUniqueId().toString().replaceAll("-", "");
+        return Regex.Replace.DASH.remove(Bukkit.getOfflinePlayer(player).getUniqueId().toString());
```
A248, line 248: Same here. String#replace replaces literal strings

core/src/main/java/me/leoko/advancedban/manager/UUIDManager.java
```diff
@@ -208,7 +209,10 @@ public String getNameFromUUID(String uuid, boolean forceInitial) {
     private String askAPI(String url, String name, String key) throws IOException {
     	MethodInterface mi = mi();
         name = name.toLowerCase();
-        HttpURLConnection request = (HttpURLConnection) new URL(url.replaceAll("%NAME%", name).replaceAll("%TIMESTAMP%", new Date().getTime() + "")).openConnection();
+
```
A248, line 212: This makes the code more complicated without much measurable gain (the slow performance of the http connection will dwarf any regex by comparison). Also, you can use String#replace and you don't need regex at all

bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
```diff
@@ -238,13 +239,13 @@ public String getIP(Object player) {
 
     @Override
     public String getInternUUID(Object player) {
-        return player instanceof OfflinePlayer ? ((OfflinePlayer) player).getUniqueId().toString().replaceAll("-", "") : "none";
+        return player instanceof OfflinePlayer ? Regex.Replace.DASH.remove(((OfflinePlayer) player).getUniqueId().toString()) : "none";
```
A248, line 242: Regex isn't needed here either. You can simply use String#replace.

bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
```diff
@@ -80,7 +81,7 @@ public String getFromUrlJson(String url, String key) {
             JSONParser jp = new JSONParser();
             JSONObject json = (JSONObject) jp.parse(new InputStreamReader(request.getInputStream()));
 
-            String[] keys = key.split("\\|");
+            String[] keys = Regex.Split.VERTICAL_LINE.split(key);
```
A248, line 84: Regex isn't needed here. String#split has a fast-path for single character delimiters, including where the backslash escape character is used.

A248 (commented):
In all but a few of these cases, using regex is not necessary. Also, there's a lot of unneeded additional complexity.

core/src/main/java/me/leoko/advancedban/utils/Command.java
iGabyTM: I changed the code to use Matcher#matches instead of String#matches because the regexes are pretty big and imo is better to have them compiled.

core/src/main/java/me/leoko/advancedban/utils/Command.java
```diff
@@ -235,7 +235,7 @@
             input -> {
                 Punishment punishment;
 
-                if (Regex.ZERO_OR_N_DIGITS.matches(input.getPrimaryData())) {
+                if (Regex.DIGITS.matches(input.getPrimaryData())) {
```
iGabyTM, line 238: Here I would personally use [`Ints#tryParse(String string)`](https://guava.dev/releases/snapshot/api/docs/com/google/common/primitives/Ints.html#tryParse(java.lang.String)) from [Guava ](https://guava.dev/) (which is available on bukkit).

PR #491, opened 2021-04-20
Fix small velocity command issues

Aurelien30000:
- CommandReceiverVelocity didn't check if the command was correct.
- Null tabcompleter for systemprefs command led to NPE (thanks Yaron).


DevLeoko (approved):
Thanks for the fixes

Issue #474, opened 2021-03-02
Allow changing maxLifetime value


A248:
_What version of AdvancedBan (`/AdvancedBan`) are you using?_
 2.3.0
 
_What kind of server do you have (`Bungeecord/single server`)?_
 n/a
 
_What server version (/version) are you using?_
n/a

_Please provide the EXACT steps required to reproduce the problem..._
1. Use a connection to a remote server
2. Impose some network or infrastructure limit on the connection lifetime
3. See the connection expire

_Any error/log post it through [pastebin.com](http://pastebin.com) and link it here. (Also include /plugins/AdvancedBan/logs/latest.log)_
https://pastebin.com/6Je0VipE

_Add any additional information below._
A couple of users have run into this bug. I'm reporting this information on their behalf.

This is both a bug report and a feature request, since to solve the bug requires the maxLifetime setting to be changeable.

**Solution**
Add a config option and use HikariConfig#setMaxLifetime to implement it.
```yaml
# Don't touch this unless you know you need to change it
# Controls the maximum time, in milliseconds a connection in the pool can stay alive
connection-max-lifetime: 1800000
```

PR #492, opened 2021-04-25
Fix issue 474

Woodie-07:
- Fixes https://github.com/DevLeoko/AdvancedBan/issues/474
- Adds a Max-Lifetime config option which is used as the max lifetime for a MySQL connection.


DevLeoko:
Please use the development branch as a target for your merge. The master branch is only for currecntly deployd version.

A248 (approved):
LGTM.

The config option isn't in UpdateManager, but I think that's fine, as this doesn't necessarily need to be added to past configs. (AdvancedBan probably should switch to a more robust config update solution than UpdateManager, but I digress)

DevLeoko (approved):

PR #498, opened 2021-05-22
Hex messages support

Homosanians:
This enhancement allows the use of hex colors other than the usual ones.

image
image

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -64,5 +64,11 @@
             <!-- Usually provided by the server implementation -->
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>net.md-5</groupId>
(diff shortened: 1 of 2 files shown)


A248:
The core of AdvancedBan cannot depend on bungeecord-chat. Otherwise, it'll break in other environments which don't include that dependency.

DevLeoko (changes requested):
See comment above for required changes

Hopefuls:
Closing due to age. Refer to previous comments by A248 and Leoko:

> The core of AdvancedBan cannot depend on bungeecord-chat. Otherwise, it'll break in other environments which don't include that dependency.

Unless theres a consistent way to implement this to be supported under non-bungeecord (aswell as keep older versions supported).

PR #499, opened 2021-05-24
Misc dependency updates

crafter23456:
I don't know if I'm doing something wrong or broken cause I'm not an professional.


A248:
HSQLDB 2.6.0 requires Java 11. Unrelated to this, it's probably a good idea for AdvancedBan to start requiring Java 11. Will need to be approved by Leoko though.

DevLeoko:
Sorry for the late answer and thanks for the PR. But about 25% of our user base is still on Java 8 so we still want to keep supporting it. (It was at 50%+ a year ago, so maybe sometime in the future)

PR #521, opened 2021-08-07
Update md_5-snapshots repo url to use https

Iaiao:
Maven blocks HTTP requests by default since 3.8.1 release.


DevLeoko (approved):
Thanks for the changes

PR #522, opened 2021-08-11
Added isSilent() method for Punishment class

### LINKED ISSUE #416: isSilent boolean
AB version: 2.2.1

Server Software: Paper

Server version: 1.16.3

Hello, i would like to suggest adding isSilent boolean, when player is banned by an operator we can add **-s** to hide the ban.

OopsieWoopsie:
No breaking changes, production-ready code :smile: 

Closes #416

### DIFF
--- a/core/src/main/java/me/leoko/advancedban/utils/Punishment.java
+++ b/core/src/main/java/me/leoko/advancedban/utils/Punishment.java
@@ -23,11 +23,16 @@ public class Punishment {
     private final String name, uuid, operator, calculation;
     private final long start, end;
     private final PunishmentType type;
+    private final boolean silent;
 
     private String reason;
     private int id;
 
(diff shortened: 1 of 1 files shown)


OopsieWoopsie:
#416

DevLeoko:
You are only saving whether the punishment is silent or not in the object attributes but are not persisting it to the database.
This results in all punishments being shown as non-silent once the server is restarted.

You will not only need to persist it to the database but also create some sort of conversion process for old databases as Aurel  mentioned in #416

PmzHeroV2:
what about adding it for the event if its not possible?

OopsieWoopsie:
The purpose of this is knowing wheter a punishment is silent at the time it is created so you can know if from an event, but what you said makes sense, this is no coherent when a punishment is get from the database because isSilent will be false in that case.
Adding isSilent to the event and not in the Punishment class would be the way :+1:

PR #523, opened 2021-08-12
Added isSilent() method for PunishmentEvent

OopsieWoopsie:


OopsieWoopsie:
The "create" method of the Punishment class calls the MethodInterface.callPunishmentEvent() method and passes the silent flag to it, from what I understand a developer is not supposed to use AdvancedBan's API and call "callPunishmentEvent", because it is supposed to be called when a Punishment is created the correct way (with the "create" method).

DevLeoko (approved):
Looks good

PR #527, opened 2021-08-29
Implement configuration option to allow changing the "CONSOLE" operator

### LINKED ISSUE #526: How can I change the "Console"
 2.3.0
 
 Single
 
1.8.8

"You have been banned by %OPERATOR%"  (Instead of OPERATOR, the person who gives the punishment writes). 
It says "You have been banned by Console" when someone automatically bans.

How can I change the "Console"?

Hopefuls:
This PR includes the following changes:

Implemented a new configuration option to allow changing the `CONSOLE` operator for when a player gets punished through the console.

- added the `ConsoleName` configuration option in the config.yml that can be changed if the server admins wish to have a different operator name when a player gets punished through the console.
- Added the method `isConsoleSender(Object sender)` to the MethodInterface to allow checking if the provided sender object is a ConsoleSender from the corresponding server environment. 
- Updated the PunishmentProcessor to include the changing of the operator when the CommandSender is a console.
- Included a failsafe if the operator name exceeds 16 characters (max length of a minecraft username) to avoid database issues.

This was tested in the following versions of minecraft:
- Spigot 1.12.2 (AdvancedBan on spigot)
- Spigot 1.12.2 + Bungeecord (AdvancedBan on bungeecord)

This should work on Spigot, aswell as Bungeecord

### DIFF
--- a/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
+++ b/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
@@ -269,11 +269,6 @@ public class BukkitMethods extends AbstractMethodInterface<YamlConfiguration> {
     }
 
     @Override
-    public boolean isConsoleSender(Object sender) {
-        return sender instanceof ConsoleCommandSender;
-    }
-
-    @Override
     public String parseJSON(InputStreamReader json, String key) {
         try {
             return ((JSONObject) new JSONParser().parse(json)).get(key).toString();
--- a/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
+++ b/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
@@ -17,6 +17,7 @@ import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.OfflinePlayer;
 import org.bukkit.command.CommandSender;
+import org.bukkit.command.ConsoleCommandSender;
 import org.bukkit.command.PluginCommand;
 import org.bukkit.configuration.InvalidConfigurationException;
 import org.bukkit.configuration.file.YamlConfiguration;
@@ -265,6 +266,11 @@ public class BukkitMethods extends AbstractMethodInterface<YamlConfiguration> {
             return true;
         }
         return false;
+    }
+
+    @Override
+    public boolean isConsoleSender(Object sender) {
+        return sender instanceof ConsoleCommandSender;
     }
 
     @Override
--- a/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
+++ b/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
@@ -17,7 +17,6 @@ import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.OfflinePlayer;
 import org.bukkit.command.CommandSender;
-import org.bukkit.command.ConsoleCommandSender;
 import org.bukkit.command.PluginCommand;
 import org.bukkit.configuration.InvalidConfigurationException;
 import org.bukkit.configuration.file.YamlConfiguration;
--- a/core/src/main/java/me/leoko/advancedban/MethodInterface.java
+++ b/core/src/main/java/me/leoko/advancedban/MethodInterface.java
@@ -251,6 +251,13 @@ public interface MethodInterface {
     boolean callCMD(Object player, String cmd);
 
     /**
+     * Check if the given Sender object is a consoleSender
+     *
+     * @param sender the sender
+     * @return the boolean
+     */
+    boolean isConsoleSender(Object sender);
+    /**
      * Get MySQL.yml file.
      *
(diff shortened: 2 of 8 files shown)


core/src/main/java/me/leoko/advancedban/MethodInterface.java
```diff
@@ -250,6 +250,13 @@
      */
     boolean callCMD(Object player, String cmd);
 
+    /**
+     * Check if the given Sender object is a consoleSender
+     *
+     * @param sender the sender
+     * @return the boolean
```
A248, line 257: I'm not sure whether Leoko is picky about this, so I'll let it be.

A248 (commented):
Code seems fine to me. I don't think a config migration is necessary -- any user coming from a previous version can still add this option in if they want to or we tell them that's how to configure the console name.

I'm not really sure about the overall concept of storing the configured console name in the database. It would be better to store whether the operator is the console in the database, and, when the punishment is displayed, use the configured console name. With that approach, the data stored in the database will not be dependent on the configuration, and updating the configuration will change the console name shown for all punishments, not just punishments made after the configuration was changed.

Other than the design approach, consider this an approval from me.

Ebubekir728:
> Code seems fine to me. I don't think a config migration is necessary -- any user coming from a previous version can still add this option in if they want to or we tell them that's how to configure the console name.
> 
> I'm not really sure about the overall concept of storing the configured console name in the database. It would be better to store whether the operator is the console in the database, and, when the punishment is displayed, use the configured console name. With that approach, the data stored in the database will not be dependent on the configuration, and updating the configuration will change the console name shown for all punishments, not just punishments made after the configuration was changed.
> 
> Other than the design approach, consider this an approval from me.

ServerType getServerType();
Cannot resolve symbol ServerType

Location: core\src\main\java\me\leoko\advancedban\MethodInterface\

Hopefuls:
> 
> ServerType getServerType();
> Cannot resolve symbol ServerType
> 
> Location: core\src\main\java\me\leoko\advancedban\MethodInterface\

This is something related to how the velocity Dev decided to implement versioning. Building it using `mvn clean package` should build it without any issues.

Hopefuls:
> I'm not really sure about the overall concept of storing the configured console name in the database. It would be better to store whether the operator is the console in the database, and, when the punishment is displayed, use the configured console name.

Yeah could take a look at that later on, just made this pr without affecting anything database-related or causing breaking changes.

core/src/main/java/me/leoko/advancedban/MethodInterface.java
Hopefuls: Tried to match the other notes from the other methods in this class, I'd be free to make any changes if Leoko wants to

DevLeoko:
Even though I agree that A248's solution would be a little cleaner (and also possible without any database migrations by just using "CONSOLE" as the constant name in the DB and changing it in the getter method) I will leave it to Hopefuls whether he wants to change that.
The only thing I think should be added in any case is an `isOperatorConsoleSender()` method as currently the checks by depended plugins are done via the operator name.

A248:
How about simply `isConsole()` for the method name?

Hopefuls:
Hey hey, back after almost a month:

I have decided to do the following things to reflect the idea from A248 and Leoko,

I have added new methods in the Punishment class called `isOperatorConsole()`, which is used by the other new method `getFormattedOperator()`.

the `getFormattedOperator()` method gets the configuration section `ConsoleName` (if present, if not, default to `CONSOLE`), which is used in the `getLayout()` method of the Punishment class aswell.

I was partially thinking about adding a new placeholder called `%BASE_OPERATOR%` or a `%CUSTOM_OPERATOR%` to return the raw, non configuration operator name (in a case of a console operator) and vise versa. So administrators could decide to for example only have the custom console name on disconnect messages and not `/banlist`.

The config-defined operator name never touches the database and just replaces the `CONSOLE` operator on the layout instead of pushing it onto the database. This allows usage on for example downgraded versions etc. Essentially allowing Legacy usage.

Feedback appreciated!

bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
```diff
@@ -17,6 +17,7 @@
 import org.bukkit.ChatColor;
 import org.bukkit.OfflinePlayer;
 import org.bukkit.command.CommandSender;
+import org.bukkit.command.ConsoleCommandSender;
```
A248, line 20:
```suggestion
```

velocity/src/main/java/me/leoko/advancedban/velocity/VelocityMethods.java
```diff
@@ -6,6 +6,7 @@
 import com.google.gson.JsonParser;
 import com.velocitypowered.api.command.CommandMeta;
 import com.velocitypowered.api.command.CommandSource;
+import com.velocitypowered.api.proxy.ConsoleCommandSource;
```
A248, line 9:
```suggestion
```

A248 (approved):
Looks good! there are a couple stray imports

DevLeoko:
Thanks a lot @Hopefuls very elegant solution 👍

PR #536, opened 2021-12-13
Added &r to Startup Messages to reset Color

herkulessi:


herkulessi:
This is Literally just there to stop the plugin from coloring The Logging Systems' info at the Beginning of each line... For some Reason this plugin did that sometimes.

DevLeoko (approved):

PR #580, opened 2022-08-15
Tab-Completion and Layout Configuration Options

Hopefuls:
This Pull-Request includes new configuration settings to disable the use of Layouts, aswell as disabling Tab-completion. Disabling Tab-completion was previously only doable by using a plugin that fully disables tab completion. 

The configuration file "config.yml" has a new section called "Use Tab Completion". This setting allows administrators to entirely disable Tab completion from AdvancedBan. This includes tab-completing users, Options and Layouts. To allow legacy configuration files, this setting is set to true by default.

The configuration file "Layouts.yml" has 2 new options called "Use Message Layouts" and "Use Time Layouts". Disabling either of them will remove Layouts from both Tab-completion, aswell as the use of it entirely. Trying to use it will respond with a message defined within the "Messages.yml" under the sections "TimeLayoutsDisabled" and "MessageLayoutsDisabled". These sections will cause issues if not present, but would not be used if Layouts are enabled by default (aka using previous configuration files). Administrators can append said sections accordingly.

As already mentioned, this should mostly not cause issues with legacy versions of configuration files, but may cause issues if "MessageLayoutsDisabled" or "TimeLayoutsDisabled" are not present. But again, this should not cause any issues through default use and can be appended by Administrators if needed.

Cheers.

### DIFF
--- a/core/src/main/java/me/leoko/advancedban/utils/Command.java
+++ b/core/src/main/java/me/leoko/advancedban/utils/Command.java
@@ -559,18 +559,9 @@ public enum Command {
 
     public TabCompleter getTabCompleter() {
         MethodInterface mi = Universal.get().getMethods();
-        if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) {
+        if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) 
             return new NullTabCompleter();
-                @Override
-                public List<String> onTabComplete(Object user, String[] args) {
-                    // Return an empty arrayList in order to not suggest tab-completions
-                    return new ArrayList<>();
-                }
-            };
-
-        }
         return tabCompleter;
-
     }
 
     public static Command getByName(String name) {
--- a/core/src/main/java/me/leoko/advancedban/utils/Command.java
+++ b/core/src/main/java/me/leoko/advancedban/utils/Command.java
@@ -18,6 +18,7 @@ import me.leoko.advancedban.utils.tabcompletion.PunishmentTabCompleter;
 import me.leoko.advancedban.utils.tabcompletion.TabCompleter;
 import org.apache.commons.lang3.ArrayUtils;
 
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
 import java.util.List;
@@ -557,7 +558,19 @@ public enum Command {
 
 
     public TabCompleter getTabCompleter() {
+        MethodInterface mi = Universal.get().getMethods();
+        if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) {
+            return new TabCompleter() {
+                @Override
+                public List<String> onTabComplete(Object user, String[] args) {
+                    // Return an empty arrayList in order to not suggest tab-completions
+                    return new ArrayList<>();
+                }
+            };
+
+        }
         return tabCompleter;
+
     }
 
     public static Command getByName(String name) {
--- a/core/src/main/java/me/leoko/advancedban/utils/Command.java
+++ b/core/src/main/java/me/leoko/advancedban/utils/Command.java
@@ -560,7 +560,7 @@ public enum Command {
     public TabCompleter getTabCompleter() {
         MethodInterface mi = Universal.get().getMethods();
         if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) {
-            return new TabCompleter() {
+            return new NullTabCompleter();
                 @Override
                 public List<String> onTabComplete(Object user, String[] args) {
                     // Return an empty arrayList in order to not suggest tab-completions
--- a/core/src/main/java/me/leoko/advancedban/utils/Command.java
+++ b/core/src/main/java/me/leoko/advancedban/utils/Command.java
@@ -531,7 +531,7 @@ public enum Command {
     private final Consumer<CommandInput> commandHandler;
     private final String usagePath;
     private final String[] names;
-    private static final NullTabCompleter nullTabCompleter = new NullTabCompleter();
+    private static final NullTabCompleter NULL_TAB_COMPLETER  = new NullTabCompleter();
 
     Command(String permission, Predicate<String[]> syntaxValidator,
             TabCompleter tabCompleter, Consumer<CommandInput> commandHandler, String usagePath, String... names) {
@@ -561,7 +561,7 @@ public enum Command {
     public TabCompleter getTabCompleter() {
         MethodInterface mi = Universal.get().getMethods();
         if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) 
-            return nullTabCompleter;
+            return NULL_TAB_COMPLETER;
         return tabCompleter;
     }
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/me/leoko/advancedban/utils/tabcompletion/PunishmentTabCompleter.java
```diff
@@ -20,6 +20,10 @@ public List<String> onTabComplete(Object user, String[] args) {
         final MethodInterface methodInterface = Universal.get().getMethods();
 
         List<String> suggestions = new ArrayList<>();
+        
+        if (!methodInterface.getBoolean(methodInterface.getConfig(), "Use Tab Completion", true)) {
```
A248, line 24: PunishmentTabCompleter is only one implementation of TabCompleter. To block all tab completion, you'd need to move this check further up the call chain.

core/src/main/java/me/leoko/advancedban/utils/tabcompletion/PunishmentTabCompleter.java
```diff
@@ -46,14 +50,20 @@ public List<String> onTabComplete(Object user, String[] args) {
...
+                    suggestions.add("#"+layout);
+                }
             }
+
         } else if((temporary && args.length == 3) || args.length == 2) {
-            suggestions.add("Reason...");
-            for (String layout : methodInterface.getKeys(methodInterface.getLayouts(), "Message")) {
-                suggestions.add("@"+layout);
+            if (methodInterface.getBoolean(methodInterface.getLayouts(), "Use Message Layouts", true)) {
+                suggestions.add("Reason...");
```
A248, line 61: I'm not sure whether "Reason..." is related to layouts in particular, or if it's supposed to represent reasons more generally.

core/src/main/resources/Layouts.yml
```diff
@@ -12,6 +11,13 @@
 # For warns you have also the variable %COUNT% which will be
 # replaced with the current amount of warns the player already received
 # Example usage: /ban Leoko @ExampleLayout
+
+# Use Message Layouts (If disabled, this will disable Tab-completion of Message Layouts aswell. Any entries within the Message section will be ignored)
+Use Message Layouts: true
+
+# Use Time Layouts (If disabled, this will disable Tab-completion of Time Layouts aswell. Any entries within the Time section will be ignored)
```
A248, line 18:
```suggestion
# Use Time Layouts (If disabled, this will disable Tab-completion of Time Layouts as well. Any entries within the Time section will be ignored)
```

core/src/main/resources/Layouts.yml
```diff
@@ -12,6 +11,13 @@
 # For warns you have also the variable %COUNT% which will be
 # replaced with the current amount of warns the player already received
 # Example usage: /ban Leoko @ExampleLayout
+
+# Use Message Layouts (If disabled, this will disable Tab-completion of Message Layouts aswell. Any entries within the Message section will be ignored)
```
A248, line 15:
```suggestion
# Use Message Layouts (If disabled, this will disable Tab-completion of Message Layouts as well. Any entries within the Message section will be ignored)
```

core/src/main/java/me/leoko/advancedban/utils/commands/PunishmentProcessor.java
```diff
@@ -83,6 +83,10 @@ private static TimeCalculation processTime(Command.CommandInput input, String uu
         input.next();
         MethodInterface mi = Universal.get().getMethods();
         if (time.matches("#.+")) {
+            if (!mi.getBoolean(mi.getLayouts(), "Use Time Layouts", true)) {
+                MessageManager.sendMessage(input.getSender(), "General.TimeLayoutsDisabled", true);
+                return null;
+        }
```
A248, line 89:
```suggestion
            }
```

core/src/main/java/me/leoko/advancedban/utils/tabcompletion/PunishmentTabCompleter.java
Hopefuls: Shouldn't be related to it as far as I'm seeing, this seems to act as a placeholder (such as the placeholder "[Name]" on the first argument or Xh Xm etc on temporary punishments). So I'd say this should still be included when disabling layouts.

core/src/main/java/me/leoko/advancedban/utils/Command.java
```diff
@@ -557,7 +558,19 @@ public void execute(Object player, String[] args) {
 
 
     public TabCompleter getTabCompleter() {
+        MethodInterface mi = Universal.get().getMethods();
+        if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) {
+            return new TabCompleter() {
```
A248, line 563: We already have an implementation of empty tab-complete:
```suggestion
            return new NullTabCompleter();
```

Hopefuls: ah perfect haven't noticed it.

core/src/main/java/me/leoko/advancedban/utils/Command.java
```diff
@@ -557,6 +558,9 @@ public void execute(Object player, String[] args) {
 
 
     public TabCompleter getTabCompleter() {
+        MethodInterface mi = Universal.get().getMethods();
+        if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) 
+            return new NullTabCompleter();
```
DevLeoko, line 563: Not that it has any actual impact on performance, but to be consistent, getter methods should only retreive attributes and not instantiate new objects. 
I would suggest creating a static final instance of a NullTabCopleter outside the method and returning it here.

DevLeoko (changes requested):

core/src/main/java/me/leoko/advancedban/utils/Command.java
Hopefuls: Done as per request, let me know if you'd want it differently 👍

core/src/main/java/me/leoko/advancedban/utils/Command.java
```diff
@@ -560,7 +561,7 @@ public void execute(Object player, String[] args) {
     public TabCompleter getTabCompleter() {
         MethodInterface mi = Universal.get().getMethods();
         if (!mi.getBoolean(mi.getConfig(), "Use Tab Completion", true)) 
-            return new NullTabCompleter();
+            return nullTabCompleter;
```
DevLeoko, line 564: Sorry, if you could just change the name from _nullTabCompleter_ to _NULL_TAB_COMPLETER_ as it is a constant. Then we will be ready to merge :)

DevLeoko (changes requested):

core/src/main/java/me/leoko/advancedban/utils/Command.java
Hopefuls: Done!

DevLeoko (approved):

DevLeoko:
Thanks for contributing 👍

PR #584, opened 2022-08-31
Wrote permission to each command

omnikeycard:


Hopefuls:
> I have also not been able to find a specification for `default: false`. Please correct me if I am wrong.

According to [spigotmc](https://www.spigotmc.org/wiki/plugin-yml/#:~:text=the%20command%20/test-,default,-Sets%20the%20default):

> default
> Sets the default value of the permission.
> If nodes does not exist, permission defaults to op.
> Allowed defaults are: true, false, op, not op
 
> true default will always grant the player the permission.
> false default will not grant the player the permission.
> op default will be true if the player is op.
> not op is the opposite behavior of op.

So as far as I am seeing, setting it to false would disable that by default, requiring the user to have one of advancedbans permissions?

Issue #589, opened 2022-10-08
AdvancedBan can't prevent chat messages even if player is muted when ChatControlPRO formatting.yml is enabled


TheJoshue:
_What version of AdvancedBan (`/AdvancedBan`) are you using?_
![image](https://user-images.githubusercontent.com/56804505/194708557-519c9818-5d77-4ebf-9f07-97c7afd46a12.png)
 
_What kind of server do you have (`Bungeecord/single server`)?_
Single server
 
_What server version (/version) are you using?_
![image](https://user-images.githubusercontent.com/56804505/194708672-b98dc28f-892e-4518-b63b-8e8f069c6aef.png)

_Please provide the EXACT steps required to reproduce the problem..._
1. Install [ChatControlPRO](https://github.com/kangarko/ChatControl-Pro/releases) plugin
2. Install AdvancedBan plugin
3. Start server to load configs
4. Go into `plugins/ChatControl/formatting.yml` file
5. Set `Enabled` to `true` to enable formatting of messages
6. Go onto server and reload CHC plugin (`/chc reload`)
7. Mute some player using AdvancedBan plugin (`/mute Player1 some reason`)
8. When that player try to send chat message he will get AdvancedBan deny message but his message will however be sent to the chat

_Any error/log post it through [pastebin.com](http://pastebin.com) and link it here. (Also include /plugins/AdvancedBan/logs/latest.log)_
No

_Add any additional information below._
No

Hopefuls:
Due to the nature of how this plugin is made, aka it not being open source, I cannot give a definite answer to that.

What I'm thinking is happening here is that ChatControlPRO does not respect the chat event status that AdvancedBan has set. Please reach out to their support instead.

Thank you.

TheJoshue:
> Due to the nature of how this plugin is made, aka it not being open source, I cannot give a definite answer to that.
> 
> What I'm thinking is happening here is that ChatControlPRO does not respect the chat event status that AdvancedBan has set. Please reach out to their support instead.
> 
> Thank you.
This is what i got as an answer from LiteBans developer - I asked him to share a part of code how he handles on player chat event if player is muted; can you please try forking and doing this what he said? because litebans successfully prevents ChatControlPRO's formatted message from being sent if player is muted; AdvancedBan not
![image](https://user-images.githubusercontent.com/56804505/194722820-6af440be-4702-48d2-be1e-d6a6402efcb7.png)

TheJoshue:
[...]

PR #591, opened 2022-10-09
Fix compatibility with some chat formatting plugins such as ChatContr…

TheJoshue:
…olPRO and ChatControlRED

### DIFF
--- a/bukkit/src/main/java/me/leoko/advancedban/bukkit/listener/ChatListener.java
+++ b/bukkit/src/main/java/me/leoko/advancedban/bukkit/listener/ChatListener.java
@@ -10,7 +10,7 @@ import org.bukkit.event.player.AsyncPlayerChatEvent;
  * Created by Leoko @ dev.skamps.eu on 16.07.2016.
  */
 public class ChatListener implements Listener {
-    @EventHandler(priority = EventPriority.HIGHEST)
+    @EventHandler(priority = EventPriority.LOWEST)
     public void onChat(AsyncPlayerChatEvent event) {
         if (Universal.get().getMethods().callChat(event.getPlayer())) {
             event.setCancelled(true);


TheJoshue:
If someone do not believe this is an issue, please consider following EXACT steps I described in my issue at https://github.com/DevLeoko/AdvancedBan/issues/589

Hopefuls:
Please refer to my previous comment mentioned on the linked issue:

This may still cause issues with other plugins that still use the `LOWEST` priority. This is due to how Spigot etc handle priorities that are set on the same level. The best way to manage this is to allow administrators to configure which priority they wish to use instead, allowing them to fine-tune and adjust according to the plugins they use.

TheJoshue:
> Please refer to my previous comment mentioned on the linked issue:
> 
> This may still cause issues with other plugins that still use the `LOWEST` priority. This is due to how Spigot etc handle priorities that are set on the same level. The best way to manage this is to allow administrators to configure which priority they wish to use instead, allowing them to fine-tune and adjust according to the plugins they use.

I totally agree, then this is a temporary solution to anyone using CHCPRO/CHCRED
Thanks.

DevLeoko:
Thanks for looking into this, but we will not be merging a "temporary fix" into the master. Especially because canceling the chat event on the LOWEST event priority would suggest that other plugins could just un-cancel the event. Which is not the case for a mute.

I see how HIGHEST priority might give some chat plugins problems if they also want to modify the event and check whether it has been canceled at the same time. So maybe we could switch to event priority HIGH for now. Could you check whether that also fixes the issue @TheJoshue ?

TheJoshue:
> Thanks for looking into this, but we will not be merging a "temporary fix" into the master. Especially because canceling the chat event on the LOWEST event priority would suggest that other plugins could just un-cancel the event. Which is not the case for a mute.
> 
> I see how HIGHEST priority might give some chat plugins problems if they also want to modify the event and check whether it has been canceled at the same time. So maybe we could switch to event priority HIGH for now. Could you check whether that also fixes the issue @TheJoshue ?

I tried with HIGH, it has the same issue when using HIGHEST - message is sent to chat despite player getting deny "you are muted" message.
I tried "NORMAL" EventPriority and it works ok, no issues.

PSYCHEER:
I have a bit another problem :D As AB is using chatlistener at HIGHEST, idk why but I need in CCR priority on MONITOR, but that...not good in long term...some update after 4 years with priority listener? while I don't use any other plugins with chatlistener, it's for me useless to have it on HIGHEST from start :D

Hopefuls:
> Thanks for looking into this, but we will not be merging a "temporary fix" into the master. Especially because canceling the chat event on the LOWEST event priority would suggest that other plugins could just un-cancel the event. Which is not the case for a mute.

PR #600, opened 2023-01-04
Add CloudnetV4 support and maven depedency fix

Phyrone:
- kick player using cloudnet v4 instant of bungeecord (like version 3 and 2 support)
  - fix player not kicked when using cloudnet v4
- updated maven dependencies necessary to compile

### CloudnetV4 is at the moment of this pull request still in release candidate state. It's still possible that the API changes, however I consider it as very unlikely.

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -56,6 +56,10 @@
                             <pattern>org.bstats</pattern>
                             <shadedPattern>${project.groupId}.shaded.org.bstats</shadedPattern>
                         </relocation>
+                        <relocation>
+                            <pattern>com.google.common</pattern>
+                            <shadedPattern>${project.groupId}.shaded.com.google.common</shadedPattern>
+                        </relocation>
                     </relocations>
                     <transformers>
                         <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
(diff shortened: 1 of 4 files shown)


DevLeoko:
Looks good - has this been tested with the latest CloudNet V4 build?

Phyrone:
I installed it on my network for a few hours now, and it works fine so far.

Phyrone:
However cloudnet v4 uses adventure text, and I was not sure first, how to convert.
I used legacy next convert for (like bungeecord TextComponent.fromLeagytext) now. I think that's the correct one.

DevLeoko (dismissed):
Thanks for the contibution

DevLeoko:
Seems like something with the Cloudnet V4 dependency is wrong (see automated test)

Phyrone:
Ok looks like i have to revert some depdency updates

Phyrone:
Looks like cloudnet 4 requires at least java 17 meanwhile advancedbans still compiles on java 1.8

Hopefuls:
Is there a solution to this? We may have to "bite the bullet" and consider AdvancedBan to bump to a newer version required. This will break backwards-compatibility, unless theres a better solution to this.

Issue #427, opened 2020-10-31
Option to Remove Notes from /history


SiuanSong:
**Feature Request**
Would be great to have a way to opt out of notes appearing in someones history.  Currently they look like punishments and have been causing some confusion with staff members.
Thanks!

DevLeoko:
Implemented by @tWizT3ddreaMr in #623

PR #623, opened 2023-07-03
FullHistory

tWizT3ddreaMr:
took a more configurable way for #427 where you can exclude any punishment type if you so choose

### DIFF
--- a/core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
+++ b/core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
@@ -134,12 +134,13 @@ public class PunishmentManager {
             putList.add(put);
         return getPunishments(target, current, putList);
     }
+
     /**
      * Get all punishments which belong to the given uuid or ip.
      *
      * @param target  the uuid or ip to search for
      * @param current if only active punishments should be included.
-     * @param putList a list of basic punishment types to search for ({@link PunishmentType#BAN} would also include Tempbans).
+     * @param putList a list of basic punishment types to search for
      *                Use <code>null</code> to search for all punishments.
      * @return the punishments
      */
--- a/core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
+++ b/core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
@@ -127,12 +127,7 @@ public class PunishmentManager {
      * @return the punishments
      */
     public List<Punishment> getPunishments(String target, PunishmentType put, boolean current) {
-        List<PunishmentType> putList= new ArrayList<>();
-        if(put == null)
-            putList = Arrays.asList(PunishmentType.values());
-        else
-            putList.add(put);
-        return getPunishments(target, current, putList);
+        return put == null ? getPunishments(target, current, PunishmentType.values()) : getPunishments(target, current, put);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
```diff
@@ -126,12 +127,17 @@ public void discard(String name) {
      * @return the punishments
      */
     public List<Punishment> getPunishments(String target, PunishmentType put, boolean current) {
```
DevLeoko, line 129: The `#getPunishments` method is part of the general API. This method may be used in other places than just by the `/history` command. So this logic should be extracted to the history command.

One idea would be to add another `getPunishments()` which accepts a list of PunishmentTypes then the history command could use that method

DevLeoko (changes requested):

tWizT3ddreaMr:
That makes sense, I made the changes based on what I interpreted of what you said.
I tried to fix the new javadoc for it, but I like i may have left it lacking
I had to rearrange the order for the new method to avoid ambiguousness since there is going to actively be nulls put in

core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
```diff
@@ -126,12 +127,33 @@ public void discard(String name) {
...
+     * @return the punishments
+     */
+    public List<Punishment> getPunishments(String target, boolean current, PunishmentType ... put ) {
+        MethodInterface mi = Universal.get().getMethods();
         List<Punishment> ptList = new ArrayList<>();
+        List<PunishmentType> putList;
+
+        if(put==null) {
+            putList = new ArrayList<>();
+            mi.getStringList(mi.getConfig(),"FullHistory").forEach((typeString -> putList.add(PunishmentType.valueOf(typeString))));
```
DevLeoko, line 149: I don't think that `put` can be null here, it is just an empty array if no args are provided or `[null]` (an array with null in it) if you pass null. Also this logic should still not be in the API method but rather in the command handler.

core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
```diff
@@ -126,12 +127,33 @@ public void discard(String name) {
...
+    /**
+     * Get all punishments which belong to the given uuid or ip.
+     *
+     * @param target  the uuid or ip to search for
+     * @param current if only active punishments should be included.
+     * @param put     an array of the basic punishment type to search for ({@link PunishmentType#BAN} would also include Tempbans).
+     *      *                Use <code>null</code> to search for all punishments.
+     * @return the punishments
+     */
+    public List<Punishment> getPunishments(String target, boolean current, PunishmentType ... put ) {
```
DevLeoko, line 142: While I really like the spread operator I would suggest keeping the same order of arguments and using an array instead. This provides more consistency.

So `getPunishments(String target, PunishmentType[] punishmentTypes, boolean current)`

core/src/main/resources/config.yml
```diff
@@ -153,3 +153,16 @@ Friendly Register Commands: false
...
+FullHistory:
+  - "BAN"
+  - "TEMP_BAN"
+  - "IP_BAN"
+  - "TEMP_IP_BAN"
+  - "MUTE"
+  - "TEMP_MUTE"
+  - "WARNING"
+  - "TEMP_WARNING"
+  - "KICK"
```
DevLeoko, line 167: I am not sure, but does it make sense to include the TEMP_XXX variants here? I think the `getPunishments()` method will automatically also fetch the TEMP variant if the base variant is included. So even if the user removes TEMP_BAN from the list, they will still be displayed as long as BAN in in the list.

DevLeoko (changes requested):

DevLeoko:
Feel free to also message me on Discord if you have questions or want to chat about the comments. 
I appreciate your contribution 👍

core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
tWizT3ddreaMr: I actually wanted to but that provided some issues with the ambiguousness since i still would be passing a null from the history I cant @notnull it

core/src/main/java/me/leoko/advancedban/manager/PunishmentManager.java
tWizT3ddreaMr: it can be if you pass a null from the command, or at least it accepts and works that way. However that is a moot point anyways as I am moving the logic to the command as requested

core/src/main/java/me/leoko/advancedban/utils/Command.java
```diff
@@ -291,9 +293,20 @@ else if(args.length == 2)
...
-                    "History", true, true),
+            input -> {
+                MethodInterface mi = Universal.get().getMethods();
+                List<PunishmentType> putList = new ArrayList<>();
+                mi.getStringList(mi.getConfig(),"FullHistory").forEach((typeString -> putList.add(PunishmentType.valueOf(typeString))));
+
+                if (!Universal.get().hasPerms(input.getSender(), "ab.history")) {
+                    MessageManager.sendMessage(input.getSender(), "General.NoPerms", true);
+                    return;
+                }
```
DevLeoko, line 304: I don't think that you need to perform a permission check here as the permission for the command is specified a few lines above

DevLeoko (changes requested):

DevLeoko (approved):

DevLeoko:
Thanks for adding this feature

Issue #557, opened 2022-02-23
[Feature Request] Notify permission for silenced punishments.


SiuanSong:
**Suggestion**
Add a permission node so staff will be able to view silenced punishments.

This would help prevent double punishments.  Currently, if I see someone doing something that requires immediate intervention with a punishment and I use -s, none of the other staff can see it.  And I can't see if anyone else did anything either, if they use -s.  This can sometimes lead to a person getting a double mute or warning for the same infraction and we wouldn't know until viewing their /history or the player alerts us to it.

DevLeoko:
Implemented by @tWizT3ddreaMr in #624

PR #624, opened 2023-07-03
Made notifications for a silent punishment

tWizT3ddreaMr:
The silent notification perm is the same as notify but silentnotify, seems like it should be that, but I was not sure how you would want to structure the perm

so permission for silent mute is ab.silentnotify.mute

This is for feature request #557 
Resubmitted for the developement branch since it wasn't letting me change base

### DIFF
--- a/core/src/main/java/me/leoko/advancedban/utils/Punishment.java
+++ b/core/src/main/java/me/leoko/advancedban/utils/Punishment.java
@@ -96,9 +96,7 @@ public class Punishment {
             }
         }
 
-        if (!silent) {
-            announce(cWarnings);
-        }
+        announce(cWarnings, silent);
 
         if (mi.isOnline(getName())) {
             final Object p = mi.getPlayer(getName());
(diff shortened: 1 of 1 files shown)


Hopefuls:
Is it possible to have this be unified with the current permissions?

Such as:
ab.silentnotify.tempban -> ab.notify.tempban.silent

tWizT3ddreaMr:
It sure is, however I did it this way so if one were to give  ab.notifify.* that wouldn't include these for ease of use and to avoid any unseen issues. 
However I am not attached to the permission being separate as most people getting one will get the other on my server.

DevLeoko:
I actually think that `ab.silentnotify.*` aligns pretty well with our current permission layout, given that we also have `ab.undoNotify.*`.
Also, I think it is good that `ab.notify.*` does not include the silent notify permissions.

So if you don't have any objections @Hopefuls I'd approve and merge

Hopefuls:
Yup looks good to me

DevLeoko (approved):
Thanks for your contribution to advancedban

PR #625, opened 2023-07-04
Found a typo for permissions

tWizT3ddreaMr:
I found like a 3 year old typo
It obviously didn't cause any issue but hey better fix it when we see it


DevLeoko (approved):
Oh, thanks for catching that 👍

PR #627, opened 2023-07-05
Update pom.xml

tWizT3ddreaMr:
Updated pom.xml to have a version that is supported, because it will not build with this branch otherwise.
I do not know how well it works, I have never run a velocity server.


Hopefuls:
I'm tbh still thinking about entirely removing velocity from the Dev branch, seeing how the Dev who implemented it isn't working on it anymore.

tWizT3ddreaMr:
If it just gets removed from the main pom then it will build and leave out in the build
While leaving the structure for someone to maybe pick up later

PR #662, opened 2025-06-29
hex color support

postyizhan:
图片

### DIFF
--- a/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
+++ b/bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
@@ -8,6 +8,7 @@ import me.leoko.advancedban.bukkit.listener.CommandReceiver;
 import me.leoko.advancedban.manager.DatabaseManager;
 import me.leoko.advancedban.manager.PunishmentManager;
 import me.leoko.advancedban.manager.UUIDManager;
+import me.leoko.advancedban.utils.ColorUtils;
 import me.leoko.advancedban.utils.Permissionable;
 import me.leoko.advancedban.utils.Punishment;
 import me.leoko.advancedban.utils.tabcompletion.TabCompleter;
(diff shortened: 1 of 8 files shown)


bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
```diff
@@ -49,6 +50,23 @@ public class BukkitMethods implements MethodInterface {
...
     private BiFunction<OfflinePlayer, String, Boolean> permissionVault;
 
+    // Check if native hex colors are supported (1.16+)
+    private static final boolean SUPPORTS_HEX_COLORS = checkHexColorSupport();
+
+    /**
+     * Check if current Bukkit version supports native hex colors
+     * @return true if supported
+     */
+    private static boolean checkHexColorSupport() {
```
Hopefuls, line 60: is this 100% viable? Can this cause issues with other plugins trying to implement this?

core/src/main/resources/Messages.yml
```diff
@@ -1,5 +1,9 @@
 General:
-  Prefix: "&c&lAdvancedBan &8&l»"
+  # Hex color prefix example: &#FF5555 is red, &#FFAA00 is orange
+  # Gradient example: {#FF5555>AdvancedBan>#FFAA00} creates red to orange gradient
+  Prefix: "&#FF5555&lAdvancedBan &8&l»"
```
Hopefuls: Honestly per default, leave with the usual color codes. The comment about supporting hex color codes is fine, as long as it is mentioned that it requires the version supporting it.

Doing so would avoid confusion with older versions not supporting it, but admins assuming it does.

core/src/main/resources/config.yml
```diff
@@ -136,3 +136,20 @@ Disable Prefix: false
...
+  Enabled: true
+
+  # Enable gradient color support ({#RRGGBB>text>#RRGGBB} format)
+  # This allows creating smooth color transitions in text
+  Gradients: true
+
+  # Fallback behavior for unsupported clients
+  # 'strip' - Remove hex colors completely
+  # 'closest' - Convert to closest traditional color (future feature)
+  # 'keep' - Keep hex codes as-is
```
Hopefuls, line 154: What would "keep" technically do? Just have them as visible characters or are they "hidden" within the text?

Edit: Fallback doesn't seem implemented yet, still interested in your idea about this

Hopefuls (changes requested):

bukkit/src/main/java/me/leoko/advancedban/bukkit/BukkitMethods.java
postyizhan: To be honest, I don't know for sure, but I don't think it will cause issues.

core/src/main/resources/Messages.yml
postyizhan: https://github.com/DevLeoko/AdvancedBan/pull/662/commits/8fa22f755153f0be176bca534ae1829eb3f266ba

